PR #63, opened 2015-03-10
Fix migrations

niekvanderkooy:
Several fixes in the migrations files that generated errors when reversing migrations (For example on a migrate:refresh)


JC5:
Thanks!

PR #67, opened 2015-04-04
Added 'encryption' config key and disabled by default 

vbali:
Encryption causes a lot of trouble when ordering lists or detecting duplicates:

- Whenever you create a new transaction and assign it to an existing category the category become duplicated (by `firstOrCreate`)
- Account and category lists cannot be ordered by name (or at least by the original name) because they ordered by encrypted values.

*Personally I prefer unencrypted values, because sometimes I would like to run direct SQL queries on the database.*

### DIFF
--- a/app/Http/Controllers/AccountController.php
+++ b/app/Http/Controllers/AccountController.php
@@ -1,5 +1,6 @@
 <?php namespace FireflyIII\Http\Controllers;
 
+use Amount;
 use Auth;
 use Carbon\Carbon;
 use Config;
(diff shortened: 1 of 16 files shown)


JC5:
Encryption was added to prevent snooping in a multi-user environment. Since I'm currently running FF3 in a multi-user environment, disabling encryption is not going to happen.

Having said that, technically, encryption is fairly useless; I can read the data if I want to, just not as easily.

Sorting and duplication are not influenced by encryption. Or rather, being able to disable them won't fix those methods; I'll have to consider some entries may be encrypted and other aren't. Whatever the case, the sort routines and ``firstOrCreate``-routines need a fix that takes encryption into account, whether your particular instance decides to use it or not.

If encryption is something that can be disabled, it'll be through an ``.env`` variable. I'll get around to it soon.

PR #69, opened 2015-04-06
Remove another iconv()

vbali:
Removed another iconv() call. The encoding issue has been fixed by grumpydictator/gchart 1.0.9.


JC5:
The gchart thing only fixes the id which got too long with my example name. I'm going to fix this one in develop, where I'm currently refactoring the GoogleChartController. Thanks!

PR #85, opened 2015-07-01
use relative path in login screen

ebbz:
I was installing firefly-III om my box. Configured my webserver to have a firefly in a sub directory. This caused some troubles when trying to log in.
The issues were caused by a leading slash in some of the links. If the slash is removed, the path will be relative and it works form the subdirectory.

### DIFF
--- a/resources/twig/auth/login.twig
+++ b/resources/twig/auth/login.twig
@@ -18,7 +18,7 @@
     <div class="login-box-body">
         <p class="login-box-msg">Sign in to start your session</p>
 
-        <form action="/auth/login" method="post">
+        <form action="auth/login" method="post">
             <input type="hidden" name="_token" value="{{ csrf_token() }}"/>
 
             <div class="form-group has-feedback">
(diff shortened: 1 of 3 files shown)


JC5:
Looks good! I can't believe I missed that.

Before I merge your pull request, could you check out what this piece of code will do on your installation?

```<form action="{{ URL.to('/auth/login') }}" method="post">```

(updated on the other pages as well of course)

It should amount to the same thing, but is more foolproof than the hard-coded URL's which admittedly, I started with myself. The next step would be to actually update the code so we can refer to these URL's by name, not by actual URL.

Let me know!

ebbz:
The URL.to method is working great. Like you said, it renders into the same result.

Would like me to submit a new pull request?

Just to know, is this syntax for twig or something? I'm not familiar with this...:)

JC5:
If you would, please! Please ignore the scrutinizer results.

The code is contained in brackets (```{{ }}```) which tells Twig to echo something. The code between the brackets is very similar to the PHP code: ```URL::to('/auth/login');```. Twig requires you to replace the ```::``` with a dot.

The code is only parsed however because [this file](https://github.com/JC5/firefly-iii/blob/master/config/twigbridge.php#L143) says "URL" is allowed in Twig.

I'll close this pull request, could you create a new one?

PR #128, opened 2015-12-11
applying demter law for account type

tonicospinelli:
- adds constants for account types


JC5:
I'm not sure if I did it right but your pull request has been pulled into develop!

tonicospinelli:
what is the flow to contributing with? what is the flux to add new feature or add improvements?

JC5:
If you see something, build something. That's the idea :P

PR #132, opened 2015-12-21
change array syntax to compliance the transifex array format

tonicospinelli:
To resources updates automatically:

- must not be short array syntax
- must be align all key-values pairs
- must not be break lines for long texts

### DIFF
--- a/resources/lang/en_US/breadcrumbs.php
+++ b/resources/lang/en_US/breadcrumbs.php
@@ -1,5 +1,5 @@
 <?php
-return [
+return array(
     'home'                  => 'Home',
 
     // accounts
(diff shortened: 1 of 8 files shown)


JC5:
I'm going to wait with this one until I hear back from Transifex. I've asked them about their array format limitations. It's not unreasonable to support the short array syntax, which has been around for _almost four years now_.

I'm not going to use wide lines because they're impossible to maintain and likewise, I feel should be supported by Transifex.

Unless they tell me it's not going to happen, I'm going to wait with merging this.

tonicospinelli:
transifex sucks!!!

PR #137, opened 2016-01-03
String is a reserved class name in php 7. 

leander091:
The upstream project has changed the class to Str.


leander091:
Sorry my bad i'm going to create a good pull request.

PR #146, opened 2016-01-18
ABN AMRO Specifix

roberthorlings:
The ABN AMRO bank procudes a basic CSV export, with different formats for the description. Moreover, the export doesn't contain any other indication of the opposing account (except for the text or iban in the description). This fix parses all known description formats available in the ABN AMRO format, and fills the opposing account properly.


JC5:
Looking good! I've merged the code into the develop-branch.

I am going to remove 'comma amount' converter. I feel that it would only lead to more problems. There are many possible formats which all would require their own converter, and this would present the user with even more options and choices.

Solutions as I see it, in order of preference:

1. Firefly is able to see which format the number is in (comma, decimal, spaces, etc). Firefly can also remove extra euro-signs, dollar signs, etc. A general "clean-up" in the amount converter combined with some smart code to match the used convention.
2. Same as above, but Firefly will assume the current locale is equal to the number's formatting. Ie. if you're using the Dutch locale, your CSV-file will have the amounts formatted as "1.000.000,12" and so on.

The problem with using locale information, whatever the source, is that it puts all these dependencies in the converters which I would like to avoid.

Oh well. I've already spotted some optimalizations which I'm going to push in a minute.

JC5:
I just realized the tag separater has two variants as well: comma and space separated. Practice what you preach! :D

roberthorlings:
I've also thought about your first solution. However, I think automatic cleanup of the amount could be complex. For example, the comma could be used (e.g. for some Dutch systems) as decimal separator, but could also be used as thousands separator. I couldn't find or think of a proper way to handle both cases.

JC5:
I know, I can't think of something decent either. Let's keep looking.

PR #148, opened 2016-01-22
Added binding for RuleGroupRepositoryInterface

roberthorlings:
This solves a bug when saving a new RuleGroup


JC5:
Lol never noticed that. I put in my rule groups manually because I was still testing. Time to finish test coverage ;).

Issue #131, opened 2015-12-20
Make reporting year configurable


webenhanced:
Hi James, I congratulate you on this nice piece of software you are sharing. Thanks.

I am wanting to add features to it and would like to know some stuff before I get going:-
- Should I always work from the 'master' branch? I ask this because I see some nice localization of number formats happening in the 'develop' branch but it is 5 commits behind 'master'.
- The first feature I want to add is making the reporting year start and end dates configurable; either from 'config' or preferences. In many countries (including my home Australia) we report for tax purposes from dates other than the calendar year. Here it runs from 1st July to 30th June.

How is the best way to communicate with you while this development is happening?

Thanks, Graham

JC5:
Hi!

First of all, thanks! I appreciate the compliment.

Always work in the "develop" branch. You can read more about the [gitflow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow/), or Google it. The best you can do is checkout the develop branch, get a sense of what you need to do and then, use Github to create a "pull request". In the pull request, you'll write all the changes you wish to implement.

As far as I am concerned, the changes you would like to implement can be as big or as small as you'd like, although I'd prefer small incremental changes, possibly in multiple pull requests.

If you want to change the reporting year, you can already select a custom range on the reports page. ``2014-07-01 - 2015-06-30``. This should work, at least partially.

As you can see on the report page, there's a whole bunch of "pre-set" ranges already. It should be fairly easy to add more ranges to that list.

To implement your idea, I would suggest something like this:

* A preference (true/false) called something like "I want to setup a custom fiscal year" on the preferences page.
* Two dates, start and end. "1st of July" and "30th of June" for example. No year selection. Or maybe this date is different every year? Who knows! Let me know. It should be put on the preferences page.
* Once this has been configured by the user, and turned on, the report page (index) could show something like this:

```
2014 2014 (fiscal year)
January 2014  February 2014 [...]
2015 2015 (fiscal year)
January 2015  February 2015  [...]
```

It would look better in HTML of course. Or maybe there's another smart way. Open an issue, write a pull request, and just add it. I would suggest building it in this order:

* Create a true/false preference on the preferences page. Make sure it can be localized and add English translations. The other ones, you can ignore them for now. Put this in one pull request.
* Create the date preferences. Put this in another pull request.
* Edit the report-index so it will respond to the preferences.

For all other stuff such as localization, bug fixes, translations and anything else, create a pull request, open an issue or do both. 

[...]

PR #149, opened 2016-01-27
Feature - custom fiscal years

webenhanced:
This feature will provide a preference to enable "Custom Fiscal Year" and ask for a start date. Many countries do not use the calendar year for tax and yearly reporting. This feature allows those people to manage their finances according to their local requirements.

Fiscal years are named using the year of the last month in the year. Thus, here in Australia, our year goes from 1st July to 30th June, and the fiscal year ending June 2016 is called the 2016 financial year.

All reports will then start their years from the first month as specified in the preference. Only month and day from the preference are used so it does not matter what year is entered.

I have created a new helper designed for fiscal functions and these are used only in the ReportHelper.

I could figure out how to create individual pull requests for each commit, so just created a clean branch from develop and added only this feature. I have tested it only on WAMP locally.

Graham


webenhanced:
This refers to issue #131 .

PR #166, opened 2016-02-02
Two factor authentication

zjean:
Hi,
I was looking into adding another layer of security to the authentication of Firefly, and decided to implement two factor authantication based on the TOTP protocol.
https://en.wikipedia.org/wiki/Time-based_One-time_Password_Algorithm

I added an option to the profile page to enable two factor auth. When enabling that, we ask the user to scan the QR code with the TOTP app, and enter the generated code.
We set a flag on the user record which we check when the user logs in. If 2fa is enabled, we redirect him (by middleware) to the page where he needs to enter the token from the app,

Maybe the pull request is quite big, please check it out if you like it.
One small bug I found is this when: When a 2fa-enabled user logs out, the Session is not fully cleared by Laravel. That means that when he logs in again, the session flag that indicates that he passed 2fa is still set.


JC5:
Alright this is pretty cool! I like the idea!

As for the PR, I'm afraid I'm going to have to turn it down. I have several reasons:

- Extending the user is cool, but I would store this with the rest of the preferences. That would also automatically encrypt the secret key. 
- You have some unrelated changes in your PR. 
- You have extended the gitignore with ide specific files. Put those in your own .gitignore_global. However, _mea culpa_, because there are plenty of ide specific files listed in the .gitignore already. I will make sure to remove those as well.
- It seems you left debug code in RedirectIfAuthenticated.
- Instead of view()->with(), use view('profile.two-factor-auth', compact('varname')). This is shorter.
- Your forms either have a direct URL as action (bad, in case they change) or no action (bad, it assumes the same URL).
- You have added the Google2FA as an alias but you don't use it.
- Your migrations break the tests.

You have added two-factor authentication as a new "layer", requiring two checks for every view. One for the login, one for the two-factor. That makes the entire application more complex because every view will check if the user is logged in AND if the user has a two-factor auth. However, this is stored in session data which isn't that secure, especially not when using cookie based sessions.

The "is the user authenticated using two factor" check could also be incorporated into the normal authenticate-middleware. Given Laravel's structure, it might be useful to build a new [guard](https://laravel.com/docs/5.2/authentication#adding-custom-guards) instead. This guard would incorporate two factor authentication and might even extend the Auth::check() command.

I'm fairly sure there examples of 2 factor auth out there.

Let me know if I can help!

zjean:
Hi,
thanks for your response. I appreciate your feedback! I don't have that much experience with Laravel, so I hadn't found the guards.
I will try to submit a new pull request with the suggestions you gave me.
Could you tell me how I can run the tests to make sure I don't break them?

Thanks!

JC5:
Cool! Remember that your code isn't lost. You might be able to re-use most of it. I would suggest waiting until after this weekend; I'm going to submit a lot of new stuff.

To make sure the tests work, you can run phpunit yourself. It should fail just like the tests did. If you follow [this link](https://travis-ci.org/JC5/firefly-iii/jobs/106550803), you get a hint as to why they failed: the migrations you wrote do not work in SQLite.

When you start on a new implementation, see if you can write a new guard. As far as I understand Laravel, a guard determines whether or not a user is logged in and why. Even if your guard doesn't really "do" anything, it can be the first step towards a fully functional two-factor-auth guard. If you submit a custom guard that doesn't do anything, I would still accept it: if you can explain why ;).

Then, start with a preference / profile page. It doesn't have to do much. Then, add functionality. Don't extend the User tables, use the preferences. This requires a slightly different approach. 

Keep PR's small, but functional. A button doesn't have to "do" anything, but it should _work_. For example, the submit button could not do anything, it shouldn't give an error message.

PR #175, opened 2016-02-08
Started over with 2fa

zjean:
Hi,
I am starting over with 2 factor auth.
In this PR I added the setting to enable 2fa in the preferences page. 
The  next step I would like to implement is to redirect the user to another page to validate the QR code if the user enables the setting.


JC5:
Looks good! Merged.

PR #176, opened 2016-02-10
Some small fixes to respect the base url

zjean:
Hi, some files did not respect the base url
Also, I like that the first input field in a form automatically receives the focus, so I added that to the guest area.


JC5:
Looking good!

Issue #107, opened 2015-08-29
Implement general "account number" instead / in addition to iban


JC5:
IBAN is too local. Add "account number".

webenhanced:
+1 for this.

JC5:
This field should be:

- In account_meta;
- Free format;
- Visible in the edit/create screen;

JC5:
- Unique over all the users accounts. Needs a validator
- Included in the CSV import (mapper, converter, the works).
- Included in the exporter

I'll build those.

PR #177, opened 2016-02-10
Tried to start with #107

zjean:
Hi,
I tried to start with implementing #107.
I added the field account number besides the IBAN, couldn't determine from the issue text the desired behaviour.
Only creating/editig an account with an account number works for now.


JC5:
Alright, I'll pull it in! Thanks!

In a commit of myself, I'm going to change the following:

- All account types may have an IBAN / account number.

I'll include the next steps in issue #107

PR #182, opened 2016-02-15
Small fix for importing amounts with comma as decimal separator

roberthorlings:


JC5:
Done. Don't add IDE specific files to .gitignore. Use your own gitignore_global file.

PR #183, opened 2016-02-15
Testing rule triggers and running rules on existing transactions

roberthorlings:
I really like the rules functionality and implemented some additional features:
* you can see which existing transactions match the triggers you entered. This allows you to easily verify whether the triggers are correct
* you can run the rules and/or rule groups on existing transactions, which can be used to update existing transactions at once.


JC5:
This is pretty neat! However, in its current form, I'm going to refuse it. Sorry man. But not all is lost!

- You've included IDE specific files in your .gitignore. Not a real biggie.
- You've included your other pull request, again. That's kind of lazy ;).
- Some of the problems you tackle have no relation to the feature: the missing ``</li>`` tags, for example.
- You're handling rules, transactions and journals directly from the controller. Now this is not a big deal because I do so myself as well. But neither of us should do that. It's either a repository or some kind of helper that should do that.
- I've spotted several TODO tags and magic numbers. "Foei!"

But, mainly, I can't help but notice you're code is pretty complex. ``testTriggers`` is pretty wild and ``executeOnExistingTransactions`` is even bigger. That's something that shouldn't be inside a controller. Plus, ``executeOnExistingTransactions`` is present in both RuleController and RuleGroupController.

I ran the "existing transactions" routine on my database. It took 5 minutes, crashed, and created almost 20MB in log. I did get a "success" flash message when I went back to the previous page, so there's that.

screen shot 2016-02-15 at 20 27 39

Although I'm not a big fan of the big flashy warning, I do like the idea of looping with a double limit. Contrary to the previous routine, this one was pretty fast. Neat! Although using do-loops makes my skin crawl. I'm not sure what would be a nice alternative but such loops always give me a "[fire and forget](https://en.wikipedia.org/wiki/Fire-and-forget)" vibe.

My suggestions for improvement:

- Offload the actual work of matching existing transactions to [a job](https://laravel.com/docs/5.2/queues). That should allow at least some users to offload this to the background. Create a job that processes 50 transactions, and creates a new job for the next 50. If a user is using "sync" they won't see the difference. Others can be told that their transactions are being matched in the background.
- Expand this routine: I would at least like to enter a date range on a separate page. Plus, adding such a second page makes sure that a user doesn't click on something by accident and completely destroys his database. I don't think the current code refuses empty triggers. Those would match everything. Or: amount > 0, amount < 100K. Chaos, I tell you!
- No double code!
- Your i18n lines are missing quotes around :title.
- No directly touching models in controllers. I know I do it too, but regardless I'm not going to stand for it. Offload this to RuleRepository (preferably for CRUD routines) or something else (try to avoid generic terms, put it under app/Rules).
- Don't ignore Eclipse files in .gitignore.
- Don't include unrelated fixes.

So you see, not all is lost. But there's room for improvement :+1:

roberthorlings:
Thanks for the elaborate reply. I agree the submitted code was still too rough and needs some refactoring. Your suggestions will help with that. I'll probably create new pull requests instead of reusing this one.

JC5:
That would be the most easy thing to do, for both of us. Small pull requests are easy to implement and force you to work "agile": whatever you submit must work, even if it does not do much. On my end, it's easy to check if it all works as it is supposed to. Small incremental changes.

Since the develop branch is not used by many people, it doesn't really matter that certain functionality does not work yet.

PR #189, opened 2016-02-17
Check for triggers that match anything

roberthorlings:
For proper implementation of a feature to run rules on existing transactions, a check was needed to see if a trigger would match any transaction. In that case, the user should be warned. To perform this check, I've extended the existing trigger objects to do type-specific checks.


JC5:
Ok the "matchesAnything" method name is kind of weird. I find it unclear. Needs a better name or a better description. But I like the idea! See also my comment about the constructor.

PR #190, opened 2016-02-17
Introduction of factories to create rule actions and rule triggers

roberthorlings:
This enables reuse of the logic of creating actions and triggers


JC5:
Looks good! I will refactor the abort() calls into new FireflyExceptions. This is picked up by the error handler.

Both actions and triggers currently _need_ a transaction journal which is obviously not necessary anymore. See also your other PR, which has ``new TJ`` instead of a better constructor.

I would suggest implementing a few static methods instead of a public constructor.

PR #191, opened 2016-02-17
New feature for testing triggers

roberthorlings:


JC5:
Nice!

JC5:
Alright this needs some work yet, massive amount of queries.

JC5:
* create a new, unsaved rule trigger
* call matchesAnything >>
* call static function getTrigger() in factory
* with itself as parameter and an empty transaction journal
* only to end up with a new trigger
* which is then used later in matchesAnything which points all the way back.

Seems a bit long

JC5:
This can be done in the validator I think.

JC5:
In testTriggers you build a routine to see if the given triggers are valid but RuleFormRequest has this test already anyway except maybe the matchesAnything routine which can be easily implemented in a validation routine.

JC5:
You forgot your changes to JournalRepositoryInterface

PR #192, opened 2016-02-17
Bugfix in parsing specific descriptions

roberthorlings:
Some specific types of descriptions used by ABNAMRO normally include a more detailed description inside. However, sometimes this description is not provided. In those cases, a detailed description is created from information that is provided.


JC5:
Just so you know, Log::debug is not going to work when you're not doing ``use Log`` at the top of the file.

JC5:
Wait it was already there, never mind! :D

PR #194, opened 2016-02-23
Implemented support to execute rulegroup on existing set of transactions

roberthorlings:
As discussed in earlier PR, it has been implemented as background job. Moreover, users are able to select a date range and a set of accounts, to decrease the set of transactions to handle.


JC5:
Looking good! Very nice :)

PR #216, opened 2016-03-16
Fix use of absolute URLs for linecharts for the report pages.

xnyhps:
The line charts on report pages weren't loading for me, turns out the graphs were trying to load the data from an absolute instead of a relative URL.


JC5:
Don't merge into master.

Issue #181, opened 2016-02-15
SSL offloading breaks


AxisNL:
I just installed firefly-iii. Looks cool! But my ssl offloader I use in front of it breaks stuff. It automatically redirects all http requests to https. My balancer/offloader talks plain http to the web server running firefly.

Firefly sees the incoming requests using http, and forces the client to go to the http version, causing issue. Can I set the url somewhere in .env?

Now, my client requests http://firefly.domain.nl/. The balancers replies with a 301 to https://firefly.domain.nl/. The client requests that, and then firefly responds with a 302 to http://firefly.domain.nl/login instead of http://firefly.domain.nl/login. All objects like style sheets e.g. also use http. If you use relative paths, I think this would not be an issue. Or I could set the hostname manually in the .env file.

JC5:
I'll do some experimentation with [this call](https://laravel.com/api/5.1/Illuminate/Routing/UrlGenerator.html#method_forceSchema).

zjean:
I experienced this issue as well, so I vote :+1:  ;-)

JC5:
Fixed.

PR #260, opened 2016-05-04
Force https schema if APP_FORCE_SSL=true in .env

zjean:
Hi,
I created a quick fix for issue #181 to force https based on an app setting in .env


JC5:
Sorry, late reply. Done!

Issue #284, opened 2016-08-05
Transaction editing broken


niekvanderkooy:
When trying to edit a transaction I get `Type error: Return value of FireflyIII\Support\Steam::phpBytes() must be of the type integer, string returned`

Probably fixable by returning `intval($string)` instead of `$string` from `FireflyIII\Support\Steam::phpBytes()`, but cannot get a dev environment up and running to investigate it (See #283)

JC5:
Good catch! I haven't run into this one myself yet. But a fix is on the way!

PR #286, opened 2016-08-05
Generalise Steam::phpBytes

niekvanderkooy:
As a follow up from #284, make room for case insensitive and abnormally large php.ini (like mine)

### DIFF
--- a/app/Support/Steam.php
+++ b/app/Support/Steam.php
@@ -220,6 +220,13 @@ class Steam
             return intval($bytes);
         }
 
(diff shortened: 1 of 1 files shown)


JC5:
There must be better methods to send gigabytes of data to your local webserver.

JC5:
:P

niekvanderkooy:
Yeahh... Reinstalled my local server a few days back - apparently this was the default php.ini xD

PR #356, opened 2016-10-12
Require PHP intl extension

telyn:
The intl extension is used for NumberFormatter at the very least. Explicitly requiring it in composer.json would make troubleshooting the install much simpler. I had to do a surprising amount of googling before finding out numfmt_create was in the intl extension.

Having done this I have a perfectly-working firefly install in a development environment though 😁

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,6 +27,7 @@
   ],
   "require": {
     "php": ">=7.0.0",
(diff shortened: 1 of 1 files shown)


JC5:
It's a good idea to include this in the composer file as well. Although the installation manual mentions this several times ;)

PR #358, opened 2016-10-20
Docker build for firefly-iii

Blackclaws:
So I've been playing around with getting firefly to reliably build and deploy using docker and docker-compose.

Right now its rather simple and can create a non-ssl apache driven container together with an additional mysql container.

If you are interested in this development feel free to pull it in. Even though I think we should make an ssl version too since this application has pretty sensitive information and deploying it with ssl should be done in production.

If you think that having the mysql database integrated into the container makes more sense that should be possible as well. I wanted to make it easily exchangeable though. 

I would however also like to allow external modification of the .env file without rebuilding the container everytime. So a future development would be moving the .env file into for example config and mounting that directory as a volume for docker then linking the .env to the base directory.

Tell me what you think.

To use this:
Create .env file first. then run docker/docker-build.sh. then run docker-compose up in the base directory

Requirements: docker + docker-compose

Hosts a port 80 localhost instance of firefly


JC5:
This is pretty cool! I will update the installation guide as well. It needs
work anyway (the guide!) . I might even build a something for Vagrant,
making it easy for people to try this (who are not familiar with docker).

I can not merge your pull request. Could you recreate it on the develop
branch? Make sure that you pull the latest changes, then create a pull
request. The master branch is for releases only :)

PR #364, opened 2016-10-26
Added dockerfile

schoentoon:
Ok, so this one is mostly based off #358. With a few difference which might actually make it easier.
First of all, this image is based of php:7-apache instead of ubuntu directly. Second of all, no configuration (.env) is stored in the image itself. To run this, use a run command that looks something like this
```docker run --name firefly-iii --link firefly-iii-db:db -v /data/firefly-iii/env:/var/www/firefly-iii/.env:ro firefly-iii```
This does assume there is a database with data already present, if not you'll have to run something like this first
```docker run --rm -it --link firefly-iii-db:db -v /data/firefly-iii/env:/var/www/firefly-iii/.env:ro firefly-iii php artisan migrate --seed --env=production```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,42 @@
+FROM php:7-apache
+
+RUN apt-get update -y && \
+    apt-get install -y --no-install-recommends libcurl4-openssl-dev \
+                                               zlib1g-dev \
+                                               libjpeg62-turbo-dev \
+                                               libpng12-dev \
+                                               libicu-dev \
+                                               libmcrypt-dev \
+                                               libedit-dev \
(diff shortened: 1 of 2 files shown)


Blackclaws:
Is there any reason why using my mechanism to determine whether data is already in the database and seeding based on that shouldn't be used?

JC5:
I just realized your pull request and this one are by different persons. I only closed the other one because it was a pull request on the master branch, which I cannot use. It's strictly develop branch.

I suggest that you edit the files and create a new pull request. As long as you use the develop branch, I can include it!

PR #545, opened 2017-01-30
remove unmatched tag

rationalboss:


JC5:
Hi! Thanks for the patch!

Unfortunately, I can't accept patches to the "master" branch. It has been reserved for new releases. You should use the "develop" branch for bug fixes and patches so they are not released immediately, but can be tested first. 

I believe this particular unmatched tag was already removed.

Thanks,

James

Issue #537, opened 2017-01-15
docker: Error response from daemon: Could not get container for firefly-iii-db.


nwwy:
After running 

    sudo docker build -t firefly-iii .

i tried to create a database with

    sudo docker run --name firefly-iii --link firefly-iii-db:db -v /data/firefly-iii/env:/var/www/firefly-iii/.env:ro firefly-iii

like described [here](https://firefly-iii.github.io/installation-guide-docker/)

But i just get this response

    docker: Error response from daemon: Could not get container for firefly-iii-db.
    See 'docker run --help'.

I get the same Error when i try to run

    docker run --name firefly-iii --link firefly-iii-db:db -v /data/firefly-iii/env:/var/www/firefly-iii/.env:ro firefly-iii

Here is my output of ``sudo docker images``

    REPOSITORY            TAG                 IMAGE ID            CREATED             SIZE
    firefly-iii           latest              60f6dbebac3e        26 minutes ago      586.4 MB
    php                   7-apache            62001e0da76d        3 weeks ago         385.7 MB
    hello-world           latest              c54a2cc56cbb        6 months ago        1.848 kB

Could you point me to the right direction here? Docker is still kind of new to me and running the commands from your installation guide didnt work after building the image.

Should i have build it with ``    sudo docker build -t firefly-iii-db .``?

JC5:
The command assumes there is a database present. Try

`docker run --rm -it --link firefly-iii-db:db -v /data/firefly-iii/env:/var/www/firefly-iii/.env:ro firefly-iii php artisan migrate --seed --env=production`

nwwy:
I already tried this and get the same Error:

    (ssh) flipp@phobos ‹ master › : ~/git/firefly-iii
    [0] % sudo docker run --rm -it --link firefly-iii-db:db -v /data/firefly-iii/env:/var/www/firefly-iii/.env:ro firefly-iii php artisan migrate --seed --env=production
    [sudo] Passwort für flipp:
    docker: Error response from daemon: Could not get container for firefly-iii-db.
    See 'docker run --help'.

JC5:
Mmm, then I don't know. I didn't supply the Docker files, I know too little about them.

nwwy:
Maybe @schoentoon knows a little bit more? But thanks for your help. =)

JC5:
Yeah I feel guilty pinging him every time people have Docker questions I hope he doesn't mind :D

schoentoon:
[...]

PR #558, opened 2017-02-11
Improve docker support

crash7:
Hello! I was having some problems trying to test this on my machine using docker and I had improved the setup a little. 

Main changes:
- Add docker-compose file
- Add environment variable to initialize the database (just change the value of INIT_DATABASE to "no" to disable it) 
- Add custom entrypoint that generates the .env file based on the environment variables (defined in docker-compose)
- Update Dockerfile (gettext-base package)

This should help solve https://github.com/firefly-iii/firefly-iii/issues/537

I can add more documentation if needed!


docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+      - FF_APP_KEY=SomeRandomStringOf32CharsExactly
+      - FF_APP_ENV=development
+    ports:
+      - "80:80"
+    links:
+      - firefly-db
+
+volumes:
+  firefly-storage:
+    driver: local
```
steelbrain: EOF file

docker/entrypoint.sh
```diff
@@ -0,0 +1,16 @@
...
+	# depends on your machine, but it may take a file to boot mysql container the first time
+	until php artisan firefly:verify &>/dev/null
+	do
+  		echo "waiting mysql"
+  		sleep 10 
+	done
+	php artisan migrate:refresh --seed
+fi
+
+exec apache2-foreground
```
steelbrain: Similarly, EOF here as well please

docker-compose.yml
JC5: Done

docker/entrypoint.sh
JC5: Done

JC5:
Merged!

crash7:
Thanks! And thanks @steelbrain for the review, I forgot the EOF since I normally use editor config and it handles this for me.

PR #560, opened 2017-02-11
Fix smooth curves around data points in graphs

elamperti:
The graphs created using Chart.js are using smooth curves that look pleasant, but overshoot or undershoot the values between data points.
We can set `cubicInterpolationMode` to `monotone` in order to avoid this behavior.

Example of what was happening: 
undershooting curve


JC5:
Nice.

Issue #561, opened 2017-02-12
docker-compose persistence


patrickkostjens:
The recently added docker-compose setup makes it a lot easier to get firefly-iii running in my opinion, which is great. However, as it currently stands it is tailored towards development only because as soon as you stop your docker containers, everything is gone. While this is very useful for development purposes, I think it would be nice to have a docker-compose setup that it is usable as a production setup (i.e. persist storage and database across restarts). It might even make sense to make this the default since this is probably used more often than the development version.

What are your ideas on this?
I would be happy to either change the current docker-compose or create a new one so it is easy to get firefly-iii running in a production ready environment.

JC5:
I'm going to ping @crash7, @schoentoon and @elohmeier, who have recently helped with these files.

You can always create pull requests (on the develop branch).

patrickkostjens:
Would you prefer a separate file, or shall I change the current one?

crash7:
Hello @patrickkostjens! 

Completely agree that we need a docker-compose for development, I was thinking on separate compose that extends de production one mounting a volume with the app, which is the main problem, but since we need also gulp for dev, we may need to use a different base image or have a different container to build that. 

If you want, we can work together on this!

patrickkostjens:
Hi @crash7 

Having a different container would make it easier to setup I think. I can't immediately think of advantages of using a shared volume between the development and production container. 

I'd be happy to work together on this. I already did a quick setup of how I see the production docker-compose file. See [this commit](https://github.com/patrickkostjens/firefly-iii/commit/1feae802c20388e911aadad31a03d2d4bff3258f). Note that this commit modifies the current docker-compose file, whereas creating a separate file for the production environemnt, so that the two can co-exist is of course also easy to do.

crash7:
Hi @patrickkostjens

I was referring to a volume that mounts the app source code into the container, in the case of the development one, not a shared volume with the production one (apart from the db). 
I saw that I your commit you made the storage folder persistent, good catch there! 

The next thing should be renaming that file to docker-compose.prod.yml, so the people that just want to use the app can get it quickly running and then creating a separate docker-compose.yml for dev (that will be also the default).

Let me know if you have a different idea!

patrickkostjens:
Hi @crash7 

A volume to mount the code for the development one sounds like a good idea. I can either add it to the current PR, or we can create a separate PR for that. That should make development a bit easier. 

For the other points, see the PR.

crash7:
[...]

PR #562, opened 2017-02-12
Use Docker volumes created by docker-compose to persist data

patrickkostjens:
As mentioned in #561 the current docker-compose setup does not persist data which means that all data will be gone when the Docker containers are stopped. This PR resolves that by creating a separate docker-compose file that is suitable for production use. It uses a Docker volume to store data for the app.

After starting (using `docker-compose -f docker-compose.prod.yml up -d`), you still need to seed the database (can't be done automatically for production AFAIK) by running `docker exec -ti firefly-iii php artisan migrate:refresh --seed`. After this, you should be ready to go and your data will be available after restarts.

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -4,6 +4,7 @@ services:
   firefly-db:
     volumes:
       - firefly-dev-storage:/var/lib/mysql
+      - .:/var/www/firefly-iii
   firefly-app:
     environment:
       - INIT_DATABASE=yes
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -4,8 +4,9 @@ services:
   firefly-db:
     volumes:
       - firefly-dev-storage:/var/lib/mysql
-      - .:/var/www/firefly-iii
   firefly-app:
+    volumes:
+      - .:/var/www/firefly-iii
     environment:
       - INIT_DATABASE=yes
       - FF_APP_ENV=development
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -1,0 +1,14 @@
+version: '2'
+
+services:
+  firefly-db:
+    volumes:
+      - firefly-dev-storage:/var/lib/mysql
+  firefly-app:
+    environment:
+      - INIT_DATABASE=yes
+      - FF_APP_ENV=development
+
+volumes:
+  firefly-dev-storage:
+    driver: local
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -5,4 +5,8 @@ services:
     environment:
       - FF_APP_ENV=production
     volumes:
-      - ./storage:/var/www/firefly-iii/storage
+      - firefly-app-storage:/var/www/firefly-iii/storage
+
+volumes:
+  firefly-app-storage:
+    driver: local
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -2,7 +2,6 @@ version: '2'
 
 services:
   firefly-app:
-    container_name: firefly-iii
     environment:
       - FF_APP_ENV=production
     volumes:
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -1,33 +1,9 @@
 version: '2'
 
 services:
-  firefly-db:
-    image: mysql:8
-    environment:
-      - MYSQL_DATABASE=firefly_db
-      - MYSQL_USER=firefly_db
-      - MYSQL_PASSWORD=firefly_db_secret
-      - MYSQL_RANDOM_ROOT_PASSWORD=yes
-    volumes:
-      - firefly-db-storage:/var/lib/mysql
-
   firefly-app:
-    image: firefly-iii
     container_name: firefly-iii
-    build: .
     environment:
-      - FF_DB_HOST=firefly-db
-      - FF_DB_NAME=firefly_db
-      - FF_DB_USER=firefly_db
-      - FF_DB_PASSWORD=firefly_db_secret
-      - FF_APP_KEY=SomeRandomStringOf32CharsExactly
       - FF_APP_ENV=production
-    ports:
-      - "80:80"
-    links:
-      - firefly-db
     volumes:
       - ./storage:/var/www/firefly-iii/storage
-volumes:
-  firefly-db-storage:
-    driver: local
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -9,7 +9,7 @@ services:
       - MYSQL_PASSWORD=firefly_db_secret
       - MYSQL_RANDOM_ROOT_PASSWORD=yes
     volumes:
-      - ./firefly-db-storage:/var/lib/mysql
+      - firefly-db-storage:/var/lib/mysql
 
   firefly-app:
     image: firefly-iii
@@ -28,3 +28,6 @@ services:
       - firefly-db
     volumes:
       - ./storage:/var/www/firefly-iii/storage
+volumes:
+  firefly-db-storage:
+    driver: local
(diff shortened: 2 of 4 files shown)


patrickkostjens:
@crash7 The reset of the database was indeed caused by the environment variable. This therefore now uses Docker name volumes for the mysql data storage. It still uses the (already existing) storage folder locally though for the app. I did this because the folder structure that is in there needs to be available for the application to work, and this folder is already set up to be used by the running application.

crash7:
@patrickkostjens we may need to tweak the dockerfile if we want to avoid that, so it will work for now. Are you planning to include the dev compose file in this PR too?

patrickkostjens:
@crash7 the development compose file is the default one (docker-compose.yml) and has not been affected by the production changes. What changes would you expect to see there?

patrickkostjens:
With the idea of a base docker-compose and overriding incorporated, the new way to start the production Dockers is now: `docker-compose -f docker-compose.yml -f docker-compose.prod.yml up -d`

docker-compose.prod.yml
```diff
@@ -0,0 +1,9 @@
+version: '2'
+
+services:
+  firefly-app:
+    container_name: firefly-iii
+    environment:
+      - FF_APP_ENV=production
+    volumes:
+      - ./storage:/var/www/firefly-iii/storage
```
crash7: Are you sure this is needed? The folder inside `/var/www/firefly-iii/` already has a full storage in it (since the dockerfile copies the complete project inside the container). I tested it with a named volume directly and seems to work fine. The only caveat is that is harder to check the app logs.

docker-compose.prod.yml
```diff
@@ -0,0 +1,9 @@
+version: '2'
+
+services:
+  firefly-app:
+    container_name: firefly-iii
```
crash7: Is this needed? Even when switching from on env to another in the same folder, docker-compose will recreate the container automatically if needed.

docker-compose.override.yml
```diff
@@ -0,0 +1,14 @@
+version: '2'
+
+services:
+  firefly-db:
+    volumes:
+      - firefly-dev-storage:/var/lib/mysql
+  firefly-app:
```
crash7: You can add more volumes here to mount the app files from your host inside the container

docker-compose.yml
```diff
@@ -13,16 +13,13 @@ services:
 
   firefly-app:
     image: firefly-iii
-    build:
-      context: .
+    build: .
     environment:
-      - INIT_DATABASE=yes
```
crash7: I would keep this here with a comment that you need to set it to "no" after the first init. You still need to run it with "yes" the first time you run the production config.

Or even better, we can use this aproach [here](https://docs.docker.com/compose/extends/#/administrative-tasks) to run tasks like _init-database_ or _compile-assets_ (gulp).

patrickkostjens:
Yes, the docs need to be updated too. In preparation for that I already commented the command for the production setup. Adding the one for development is also good. So, your comment will probably serve as a good base for the docs.

docker-compose.prod.yml
patrickkostjens: It's not required, but it does make running the artisan command easier to run since you can simply run against the `firefly-iii` container. If there are other ways to get that done though, I don't see a reason to leave this in.

docker-compose.prod.yml
patrickkostjens: When I used a volume mount, I got an error that the sessions folder did not exist in storage/framework as soon as I logged in. However, if you know a way to get it working with a Docker volume, that would also be fine. I think the volume is mounted after the files are copied (i.e. copying is a build step; volume mounting is done when running the container). Therefore, the folder might be empty when the application starts with a Docker volume.

docker-compose.prod.yml
crash7: Oh, right! You can do `docker-compose exec firefly-app php artisan <command>` and that will run it inside the app container without needing to know the name.

docker-compose.override.yml
patrickkostjens: Totally forgot about that part. I only did the splitting into base/dev/prod instead of mounting the dev files as discussed. Will fix.

docker-compose.yml
patrickkostjens: Does this command also work when the production environment is set? I am not entirely sure, but I thought I saw that not happening.

Although, even if it does: I think that for the safety of production data, it might be better to run a manual command once to get it al started. If you forget to remove this environment variable that would otherwise mean that you loose all your data, which, in my opinion, is a bigger problem.

docker-compose.prod.yml
patrickkostjens: I did not know about that one, thanks! I'll remove the container name then.

docker-compose.yml
crash7: Agree, making it a separate task sounds more reasonable than using the environment variable.

docker-compose.prod.yml
crash7: That's strange, I changed this to `firefly-app-storage:/var/www/firefly-iii/storage` and added the volume at the end of the file (like the database one) and this is the content of storage the first time run the container:
```
root@c2d7dae5bc11:/var/www/firefly-iii# ls storage/   
app  build  database  debugbar	export	framework  logs  upload

root@c2d7dae5bc11:/var/www/firefly-iii# ls storage/framework/
cache  sessions  views
```

Don't get me wrong, the current line works fine too. I always try make the compose files, in a way I would be able to just copy-paste them in a server folder and just run them from there without cloning the whole app repository.
(However, to do this with firefly-iii we also need to push the image to the docker-hub)

patrickkostjens: I'll try again. Maybe I did something wrong last time I checked.

patrickkostjens:
I believe all comments have now been discussed/incorporated.

docker-compose.override.yml
```diff
@@ -0,0 +1,15 @@
+version: '2'
+
+services:
+  firefly-db:
+    volumes:
+      - firefly-dev-storage:/var/lib/mysql
+      - .:/var/www/firefly-iii
```
crash7: Important note here, since the folder you are mounting doesn't have the _vendor_ folder, you need to run `compose install` separately to make this work.

patrickkostjens: That's actually handled by the [Dockerfile](https://github.com/firefly-iii/firefly-iii/blob/develop/Dockerfile#L45)

crash7: The Dockerfile is running it in the folder inside the container after doing a copy from your host folder, therefore the vendor folder is only inside the container, and when you mount your host folder, you override the one inside the container.

patrickkostjens: What you say sounds like what I would expect to happen, but I just ran setup, and the vendor folder was present and the app was working.

crash7: Just noticed that the line is under firefly-db service instead of firefly-app

patrickkostjens: Ah, that explains a lot. I'll fix that, but are we going to leave the `compose install` up to the user?

crash7: Just for dev, as a separate task like the migration one. The other option would be to mount the folder individually sounds really weird to me.

patrickkostjens: I couldn't think of a nice way to do it either besides doing it manually. I updated the PR with the correct mount.

crash7: As long as it is documented, it shouldn't be a problem for other devs.

JC5:
Wow, thanks yous! I'll update the documentation asap! Let me know if I can merge this pull request.

crash7:
Good to merge from my side! 👍

patrickkostjens:
I also think it can be merged.

JC5:
Let me know if these instructions are correct!

https://firefly-iii.github.io/installation-guide-docker/

patrickkostjens:
The instructions look good to me.

BTW: great job creating this project. Keep up the good work.

JC5:
Thanks, I appreciate it 👍

crash7:
Excellent doc page! I noticed that the production command has two up (my bad, sorry). The right command is:
`docker-compose -f docker-compose.yml -f docker-compose.prod.yml up`

PR #579, opened 2017-02-19
Transactions without budget should only return budgetable transactions

Vincentv92:
When I was assigning transactions to budgets I noticed that in the view there were transactions which could not be assigned to budgets (Transfers, revenue, opening balance)

This pull request improves this view, only return budgetable transactions

Tests ran fine

### DIFF
--- a/app/Helpers/Collector/JournalCollector.php
+++ b/app/Helpers/Collector/JournalCollector.php
@@ -518,6 +518,7 @@ class JournalCollector implements JournalCollectorInterface
             function (EloquentBuilder $q) {
                 $q->whereNull('budget_transaction.budget_id');
                 $q->whereNull('budget_transaction_journal.budget_id');
+                $q->where('transaction_types.type', '=', TransactionType::WITHDRAWAL);
             }
         );
 


JC5:
This is for the list at `/budgets/list/no-budget`, right?

It would have my preference to add a filter in the controller, instead of changing the collector, even though it amounts to the same thing.

That particular page needs an overhaul, come to think of it.

Vincentv92:
Yes its for `/budgets/list/no-budget`
Although this change also impacts the charts, according to my IDE, this function is there also used.

It makes sense to put that logic in the controller, however, since this function exists inside Collector, it also makes sense to let the function do what it should do. Thats why I added it there.

You mentioned it needed an overhaul, what needs to be changed to improve this? Anything in particular?

JC5:
You're right. However, I'm inclined to keep the function withoutBudget() as simple as possible, and not let it judge on what kind of a transaction should be returned. That's up to the "user", which is us.

At this moment, the no-budget-list follows the date-range that you see in the top right corner. But views like `/budgets/show/x` and `/accounts/show/x` are capable of showing a list of months, the relevant amounts and other details. Some charts even!

They are composed of three views: a default view, a "specific period"-view and a "all-of-them"-view.

This is something I want to do with the no-budget-list as well.

PR #586, opened 2017-02-22
Add development getting started.

Zsub:
This will make it easier for people to get started. The only reason I got this far is because I used Laravel myself in an earlier project.


JC5:
I'm not sure this is something for the contributing file. I'd rather put this somewhere up on the website, but I also have to say: if you don't know how to clone this repository to develop in it / with it, I'm not sure if you should be developing in the first place ;).

Zsub:
Very fair point, although more (quality over quantity of course) documentation generally never hurts. In my view these were just a few little 'gotcha's' that while I could figure them out, they might form too big of a hurdle for someone who is versed in PHP but used to Symfony, for example.

Your point that this might be something for the website stands, I'll see if I can make a PR to clarify the installation a little more.

PR #587, opened 2017-02-22
Fix saving a tag’s date

Zsub:
The `date` function takes the fieldname where a date is stored, not the literal date.

### DIFF
--- a/app/Http/Requests/TagFormRequest.php
+++ b/app/Http/Requests/TagFormRequest.php
@@ -45,7 +45,6 @@ class TagFormRequest extends Request
             $longitude = null;
             $zoomLevel = null;
         }
-        $date = $this->get('date') ?? '';
 
         $data = [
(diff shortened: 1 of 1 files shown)


Zsub:
Just to show I can do something else than just nag ;)

JC5:
This time I can nag ;).

If you remove the line that declares $date, (`$date = $this->get('date') ?? '';`) the pull request is perfect. The date() function already does this so it's superfluous.

JC5:
But let me say, nice find!

Zsub:
Do you prefer that in one or two commits?

JC5:
Oh that's not important. The last time CrowdIn (localization) pushed something it was 120 single commits with just a newline removed.

Zsub:
Voila. Side effect is that loading tags now also works, including showing the nice little map :)

PR #594, opened 2017-02-24
Show suggested monthly savings for a piggybank

Zsub:
If a piggybank has both a target date and a target amount, show how much money needs to be added to the piggybank each month to achieve both targets.

Strings are currently hard-coded because I want to gauge the reaction to this :)

### DIFF
--- a/app/Models/PiggyBank.php
+++ b/app/Models/PiggyBank.php
@@ -129,6 +129,17 @@ class PiggyBank extends Model
 
     }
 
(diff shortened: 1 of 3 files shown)


JC5:
Like!

JC5:
Yes, I like it! I've merged your pull request. Feel free to add translatable strings and/or other changes in a second PR.

I just realised I could have left this one open but meh 😋

PR #598, opened 2017-02-25
Ignore IntelliJ module files

Zsub:


JC5:
You've actually reminded me that it's best practice to remove IDE specific files from the .gitignore file and put those in your own global file.

So I'm going to close your PR instead, and clean up the file. Because I just noticed it also contains my own IDE's file references.

Zsub:
Huh, TIL.

PR #606, opened 2017-03-01
Set default piggy for new transaction

### LINKED ISSUE #605: Piggybanks with single quotes in their name get filled from withdrawals
That may not the the most descriptive title ever.

It is possible to get the amount left for piggybanks below zero. I have verified this on the demo site. Steps are:

1. Fill all existing piggybanks for a savings account, so there is zero money left for new piggybanks.
1. Create a new piggybank for that savings account with a single quote in its name (such as `'puter`, because you have just seen the LEGO Batman movie), with a target amount (such as €100).
1. Create a new withdrawal from the checking account, for example for €10.
1. Observe that there is now -€10 available for piggybanks and the `'puter` piggybank contains €10.
1. Rename the piggybank (for example to `Computer`).
1. Create another withdrawal, which does not get deposited in the piggybank.

Findings from (ongoing) debugging:

- It is not possible to reproduce this for piggybanks linked to the same account as the withdrawal.

Zsub:
Fixes #605. By explicitly setting the selected piggybank to the 0-piggy, new transactions will not inadvertently get coupled to a piggybank if the piggy’s name starts with characters that get sorted before `(` (such as `!` or `’`).

### DIFF
--- a/resources/views/transactions/single/create.twig
+++ b/resources/views/transactions/single/create.twig
@@ -77,7 +77,7 @@
                         {{ ExpandedForm.text('tags') }}
 
                         <!-- RELATE THIS TRANSFER TO A PIGGY BANK -->
-                        {{ ExpandedForm.select('piggy_bank_id', piggies) }}
+                        {{ ExpandedForm.select('piggy_bank_id', piggies, '0') }}
 
                     </div>
                 </div>


JC5:
Nice catch :+1:

PR #663, opened 2017-06-06
Add pgsql dependencies to Dockerfile

kressh:
Hello. This tiny change makes possible to use firefly on pgsql driver in docker container.

PR #664, opened 2017-06-06
Add pgsql dependencies to Dockerfile

kressh:
fwd #663

> Hello. This tiny change makes possible to use firefly on pgsql driver in docker container.


JC5:
Cool!

PR #710, opened 2017-07-19
build, scrutinizer and coverage badges to the top

simonsmiley:
I had a hard time finding this information, I guess most would expect to find it in the top along the other badges


simonsmiley:
oops, used the wrong base branch and github somehow added other commits to it. I'll open new merge requests

PR #711, opened 2017-07-19
Move build, scrutinizer and coverage badges to the top

simonsmiley:
I had a hard time finding them.  
I guess most people would expect them next to the other badges


JC5:
I'm sorry, but the badges stay at the bottom, as much as I like them it feels cluttered in the top.

PR #712, opened 2017-07-19
Make charts easier to read

simonsmiley:
* Do not fill the area below the lines
* Provide border color
* Decrease transparency of border


JC5:
I might dial it back to 0.4 but nothing I do not like. Thanks!

PR #807, opened 2017-09-02
Fix reseting action type when editing rule

### LINKED ISSUE #799: Action resets when editing rule
When editing existing rule action always resets to the first option - "Set category to...".

See "pkoziol test rule" on demo site. List of rules says that rule will add tag "pkoziol test", but edit form has selected "Set category to..." and action value set to "pkoziol test".

pkoziol:
Fixes #799


JC5:
Nice, merged!

PR #808, opened 2017-09-02
Fix search that worked only for lowercase queries

### LINKED ISSUE #798: Search works only for lowercase
Noticed on my installation, but also happens on demo. Searching for "[Rent](https://firefly-iii.nder.be/search?q=Rent)" results in 0 transactions, but "[rent](https://firefly-iii.nder.be/search?q=rent)" results in 20 transactions that all have "Rent" capitalized.

pkoziol:
Fixes #798


JC5:
Nice, merged!

PR #824, opened 2017-09-09
Docker: Use working directory instead of hardcoded dist version

DieBauer:
- By using the current (.) working directory, we can get the latest changes in the Docker image.
  If the building of Docker images is automated, it will automatically pick up the checked out version.
- Rearrange some Docker layers things that don't invalidate that much.


JC5:
Nice!

PR #855, opened 2017-09-17
Enable mod_ssl in docker container

ms32035:
While Apache configuration file will still have to be customize, this makes one file less to modify when running Docker containers


JC5:
Nice! Merged!

Issue #105, opened 2015-08-23
No Index, No Follow


xzaz:
Could you add a No-Index No-Follow meta data to the header of the login page? Or a option to do so?

JC5:
Sure thing! In the mean time, you could add a robots.txt yourself. I'll get on it.

JC5:
Errr, I mean update it. It's already there.

PR #878, opened 2017-09-30
fix typo in nofollow

Findus23:
Hi,

Refs #105 

Just noticed the typo and thought I should fix it before someone wonders why it isn't recognized (even though it shouldn't matter)


JC5:
Hah, nice find! Thanks for fixing this. Merged!

PR #961, opened 2017-10-30
Docker-Compose Point To Dockerhub Builds

alazare619:
Fixes # (if relevant)
Create a Standard Docker-Compose that uses the dockerhub builds. 
Changes in this pull request:


alazare619:
Note Port is set to 9001 Amend if necessary. 

All a user needs to run firefly-iii is just this one file now.

Create a directory. Place the one yaml file in it.

Then docker-compose up -d

After its ran seed the database with the docker-compose exec 3 commands and its all done nothing else is needed. if they want to update just docker-compose down then docker-compose up -d again and it will update :)

PR #1055, opened 2017-12-18
Update README.md

simhnna:
Fix the url for the demo account

I know this is against master and not dev, but it's only the readme that should be changed.
the current url is not reachable

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ There are many financial reports available.
 
 ## Want to try Firefly III?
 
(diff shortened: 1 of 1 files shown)


JC5:
Thanks for noticing so quickly!

I changed to a new hosting provider today. I’ve recreated the sub domain and it should work again.

I’m going to close this one. The demo site should work again within 24 hours but probably within the hour. 

Thanks again, I appreciate your input :)

PR #1078, opened 2017-12-30
Introduced bulk edit endpoint for bulk edit of categories and tags

### LINKED ISSUE #509: Bulk edit
Hi, I would find it very helpful to have something like a bulk edit for transactions, where the user should be able to change values for all selected transactions.
So if I select a number of transactions and click the non-yet existing bulk edit button, change f.e. the category just once and click update -> all selected transactions should get the new value.

This would be helpful after big imports or if one does add an new category and want to change the accounts or category for a lot of old transactions.

vicmosin:
Fixes #509

Changes in this pull request:

- Allow to bulk edit category and tags while mass edit transactions

@JC5
In  addition to conversation in issue: I am agree that rules should be used during the import process, but sometimes, after a period of time, you realize that some of the transactions were imported with wrong category/tags or you just want to upgrade your finance structure by introducing new category/tag. And reimporting doesn't sound a proper way to do on-fly changes.


JC5:
Hi Victor,

This is very nice! I like your idea a lot. I shall integrate the changes into the develop-branch of Firefly III. I'm not sure yet how, I see there is a conflict and I am not sure how to resolve it. The commands they suggest give me errors, so I may have to try something else.

If it fails, I might just copy your changes and submit them to the develop-branch myself. Please let me know if is that OK with you. You would lose the commit-credits, I think.

There are some things I would do differently myself. Is it OK if I continue with your work?

In any case, bulk-edit will be part of the next release :+1:

Cheers,
James

vicmosin:
I fixed the conflicts, sorry for that.
You can use the code how ever you want:)
Could you please point me to your release plan so I know when to expect official release?

JC5:
No problem! There's not really a release track, except for the milestones. I should start working on something like that :D

vicmosin:
So when next release will be?

JC5:
End of January, v4.7.0

pkoziol:
Thanks for the feature @vicmosin!

Just one suggestion: checkboxes "Don't update X" should be checked by default to avoid accidentally overwriting transactions with empty values. I'm assuming that more people will want to change just category or budget or tags than change all three at once. However, to avoid apposite situation (need to uncheck checkboxes every time you want change value), they can automatically uncheck on input's onChange event.

JC5:
I’ve been fiddling with his code and there are some changes on the demo site.

pkoziol:
I was using demo site just before posting my comment. Checked now and I don't see any changes - checkboxes are still unchecked by default.

BTW, I got notification from Crowdin about new strings to translate and wanted to see them in action to understand context, but I was too fast and demo site wasn't updated yet. Is there any way to check exact commit that was deployed on demo site?

And two more suggestions/bugs:

1. Cancel button navigates to the index page instead of previous page as someone (at least me) would expect. Same in old mass-edit (Edit a number of transactions). Maybe you could use JavaScript and `History.back()` to avoid checking referer header or passing return URL in query string? It would have additional advantage that browser would rember which checkboxes were checked.

2. When navigating to the previous page using browser's back button, checkboxes used to select transactions are hidden. Chrome remembers that they're checked, so I guess that it should be possible to detect that from JavaScript on document load and automatically show them when at least one is checked. [GIF showing what I'm talking about](https://gfycat.com/DisgustingUncomfortableBasilisk

JC5:
Ah no those changes aren't implemented yet, I wasn't sure which code you were seeing :).

1. Yes, agreed.
2. Unfortunately, I cannot reliably detect if a user is coming back from a previous page. Javascript trigger _may_ fire but they also _may_ not.

PR #1093, opened 2018-01-04
create .htaccess file

jinformatique:
Changes in this pull request:

- To force HTTPS
- To hide directory listing
- To prevent access to .env and other files

@JC5

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,0 +1,12 @@
+# To force HTTPS
+RewriteEngine On
+RewriteCond %{HTTPS} off
+RewriteRule (.*) https://%{HTTP_HOST}%{REQUEST_URI} [R,L]
+
+# To hide directory listing
+Options All -Indexes
+
+# To prevent access to .env and other files
+<Files .*>
+    Deny from all
+</Files>


jinformatique:
I hope this makes sense. I noticed that someone could access my .env file simply going to the URL:
https://<website.com>/firefly-iii/.env

JC5:
Nice! I see a few things I would do differently rhough:

1. The change-over to https is useful but should be commented out by default. A lot of people use reverse proxies which do not support https by default.
2. Shouldn't it be enough to redirect the user to /public/ using a Permanent Redirect? If people put the entirety of Firefly III into a publicly accessible directory (this is a bad idea btw because all your fils and session data are open to the world as well) then they should always end up in /public/ if they visit `/`.

I've updated your pull request to go to the `develop`-branch.

jinformatique:
Yes ok, the thing is to test if visiting /firefly-iii/.env does the redirect to /public/ as well.
The main point is to avoid revealing the .env file to internet.

JC5:
Yes, if it gives a 403 it's OK.

Keep in mind that your `storage`-directory is also open to the internet (with or without htaccess file) and this is not good.

PR #1097, opened 2018-01-06
Fix issue with accounts in select map not matching

kelvinhammond:
When mapping Opposing Accounts the select input's values are incorrect and do not match account ids.

Fixes # (if relevant)

Changes in this pull request:

- 
- 
- 

@JC5


JC5:
Thanks! I shall do some extra testing. The other mappers will suffer from this as well. I hope to have it fixed today.

PR #1107, opened 2018-01-08
Fix charts in Internet Explorer

devlearner:
As it stands, I happened to notice that the charts in Firefly-III don't render on IE.  
Turns out it seems like IE has no support for arrow function expression (and thus throws a syntax error).

Fortunately, the fix seems to be quite a simple one.    Once I changed the particular `=>` instance at L91, charts.js, the charts seem to be back again on IE.    So basically this patch's to make IE happy.

Tested it out on my machine, 4.6.13 with this patch, and both Chrome and IE 11 successfully render the charts.

### DIFF
--- a/public/js/ff/charts.js
+++ b/public/js/ff/charts.js
@@ -88,7 +88,9 @@ const verticalLinePlugin = {
 
     afterDatasetsDraw: function (chart, easing) {
         if (chart.config.lineAtIndex) {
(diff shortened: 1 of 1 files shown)


JC5:
Nice!

JC5:
The demo site has been updated.

devlearner:
Wow! Indeed.  Many thanks (as always) for your help and fast response 👍

PR #1145, opened 2018-01-23
Include Woff fonts as well

devlearner:
... for ~IE~ browsers support

(well... to be fair, it also seems to cover some slightly older Android browsers and Safari on earlier Mac OSs; aside from IE)

hope that you won't mind shipping a few more font files, but it might hopefully help in bringing a more unified experience across browsers 
(I've tested Firefox, and it seems, probably also other modern browsers, to be smart enough to load the woff2 files without requesting again for the woff files, so the negative impact would probably be small in including both font formats)

I've taken the font files from [google-webfonts-helper](https://google-webfonts-helper.herokuapp.com/); and it suggests woff and woff2 would probably be able to cover all "Modern Browsers".

As an aside I've also updated woff2 files for Lato and Roboto fonts (I didn't compare the versions, but judging from the last commit date I reckon a newer version has probably been available).

Versions of the woff font files just for the record:

| Font                        |      Version                                       |
| --------------------  |  ------------------------------------- |
| Source Sans Pro     |      Last modified 2017-10-11 (v11) |
| Lato                        |      Last modified 2017-10-11 (v14) |
| Roboto                   |      Last modified 2017-10-16 (v18) |

p.s.   Just curious, for Source Sans Pro I happen to notice that regular and italics typefaces are included for the various weights (300, 400, 600, 700) except that 700 has only got regular (without italics) typeface, was that intentional?

i.e.

|   Font weight  |   Typefaces included   |
|  --------------  |   ---------------------- |
|   300  |  Light, Light Italic   |
|   400  |  Regular, Italic   |
|   600  |  SemiBold, SemiBold Italic |
|   700  |  Bold    |

in order words, there seems currently to be no Bold Italics typeface for the font.
I think it probably isn't currently used.   But for completeness I would be happy to include Bold Italics as well if you wish (just let me know).

Thank you


JC5:
This is pretty nice! I would appreciate it if you could include Bold Italics as well. It seems you're all geared up to do so :wink:

Thank you, I appreciate the effort!

devlearner:
Thanks, James!
(actually this' pretty minor stuff indeed; hope that when I could code better I could contribute more code-wise&mdash;where you've been doing an excellent job 👍 )

JC5:
Thanks, I appreciate the compliments! Merged your commits, it's live on the demo site. :+1:

PR #1184, opened 2018-02-20
Add Docker support for sqlite

UnrealKazu:
Changes in this pull request:

- Extra environment variable `FF_DB_CONNECTION` can now be used, and will be reflected in the `.env` file.
- All DB variables not used by sqlite will be removed from the `.env` file
- An sqlite database file is created if there isn't one available

@JC5

There is no open issue for this, but I'm a big fan of Docker and I saw that even though Firefly itself supports sqlite, the Dockerfile did not. This is a small extension to make this possible. Any other databases should still be working, and it is even possible to switch between database engines if one wants to (because the `.env` file is still built from scratch on each boot).

I'm already using this in my own local environment, and it works great. Should you accept this pull request, I will also update the Docker entry in the docs to reflect this new feature.

### DIFF
--- a/.env.docker
+++ b/.env.docker
@@ -19,7 +19,7 @@ TRUSTED_PROXIES=${TRUSTED_PROXIES}
 
 # Database credentials. Make sure the database exists. I recommend a dedicated user for Firefly III
 # If you use SQLite, set connection to `sqlite` and remove the database, username and password settings.
-DB_CONNECTION=mysql
+DB_CONNECTION=${FF_DB_CONNECTION}
 DB_HOST=${FF_DB_HOST}
 DB_PORT=3306
 DB_DATABASE=${FF_DB_NAME}
--- a/docker/docker.env
+++ b/docker/docker.env
@@ -1,5 +1,6 @@
 FF_APP_ENV=local
 FF_APP_KEY=S0m3R@nd0mString0f32Ch@rsEx@ct1y
+FF_DB_CONNECTION=mysql
 FF_DB_HOST=
 FF_DB_NAME=
 FF_DB_USER=
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -9,6 +9,10 @@ cat .env.docker | envsubst > .env
 # if the database is an sqlite database, we need to remove all non-connection variables
 if [ "$FF_DB_CONNECTION" == "sqlite" ]; then
     sed -i '/DB_/{/DB_CONNECTION/!d;}' .env
+
+    # make sure own the sqlite database file
+    chown -R www-data:www-data $FIREFLY_PATH/storage/database
+    chmod -R 775 $FIREFLY_PATH/storage/database
 fi
 
 cat .env
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -10,6 +10,11 @@ cat .env.docker | envsubst > .env
 if [ "$FF_DB_CONNECTION" == "sqlite" ]; then
     sed -i '/DB_/{/DB_CONNECTION/!d;}' .env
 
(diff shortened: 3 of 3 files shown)


JC5:
Sorry, but no. This makes a docker instance more difficult to upgrade and violates the single responsibility model. Plus, SQLite is slow and gives Firefly III a bad user experience already. 

A docker instance should be easily upgradeable and replaceable and this would prevent both.

UnrealKazu:
That's a bummer. I am aware of the limitations of sqlite, but my intentions for this PR were to give means for those who want to try out your application (outside of the demo env.) without needing a full blown SQL database, or for those who simply don't want to.

Still, I'm interested in your motivations behind your arguments. For the sake of possibly improving my contribution and getting to know what your thoughts are behind them.

In terms of upgradability: the addition of the `DB_CONNECTION` variable is set up in such a way that it should work on any existing Docker environment, due to the variable having a default `mysql` value. PostgreSQL isn't supported through Docker, so all existing environments have to be MySQL only. Hence anyone who would upgrade to this version of the application would not need to change anything to their run commands or compose file.

Regarding the single responsibility, I agree that the container should only focus on providing the application itself, and to not care about any storage. It was my intention that any user who would be using sqlite would *have* to mount the `storage/database` volume, otherwise all data would be dropped when the container is removed. I see now that having the container create an sqlite file when the `DB_CONNECTION` is set to `sqlite` during bootstrap would mess with the single responsibility principle, so maybe that should be removed. This would force the user to mount his own sqlite database, which would result in more or less the same functionality as one would get when using MySQL.

All of this would of course be irrelevant if you rather not support sqlite at all, due to its limitations causing a negative experience for any user. But in my opinion this PR would still maintain the ease of upgrade and replaceability of any Docker instance. Should I've made any false assumptions, please say so!

JC5:
Hey, thanks for your reply!

I believe that docker makes it equally easy to spin up MySQL or SQLite. Another docker instance is not that much trouble. As such I much prefer MySQL. 

My experience so far has been that people who want to test Firefly III spin up the bare minimum of requirements, appreciate the app but change nothing. They leave me dealing with their crappy environments. They have a tightly coupled badly configured docker instance built in 2014 with everything neatly built in and if I can’t help them upgrade it? No. 

I would much rather force them to do it right the first time making it easier for everybody involved, including me. 

The same argument can be made for Apache versus nginx, Postgres and MSSQL, and other exotic combinations that the docker image could support. 

Docker gives me the opportunity to do it right once for many people. As such the less choice they have the better. And of course feel free to build something with SQLite or Postgres or nginx and I will gladly help you solve any issues you might have. 

It’s just not something I want to offer out of the box. 

I know that, given the right combination of configuration SQLite can be easy to use, backup and install. A single file in persistent storage next to all the other files is easy, can’t disagree with you there. 

However SQLite and Laravel are prone to fight amongst themselves when parsing data. Dates, amounts and other values tend to be less exact in SQLite. All of this can be solved by Firefly III (and it is) but I like to avoid it when I can. 

Since I have the choice I would much rather have everybody use MySQL.

UnrealKazu:
Thanks for your detailed elaboration! I can completely understand the desire to force users to use the application in such a way whereas you can safely say that it works as intended. Especially if the experience would be negatively affected when using sqlite, not only for the user but for you as well when being faced with complex bugs due to the nature of the database engine.

I wasn't aware that people would actually try to make Docker environments where they would try to cram as much functionality and applications as they desire into one Docker container, completely overlooking the fact that this is not the way Docker is intended to be used. With this in mind, and now that I understand what your motivations are, I no longer see any use in this PR and as of such will close it.

Thanks again, I will continue to look into (and use) your application. Hopefully I can contribute in some other way in the future.

PR #1211, opened 2018-03-01
3 new european currencies in database seed

m0nhawk:
Added 3 new currencies for european counties: Ukraine, Poland, Turkey.


JC5:
Nice! :+1:

I feel obliged to say that I'm not going to add all currencies of the world, so this will be all I think :).

PR #1254, opened 2018-03-14
Add support for other database types

Obbi89:
Changes in this pull request:

- Allows users to use sqlite with docker (see sample compose file below)

```yaml
version: "3.2"
services:
  firefly_iii_app:
    environment:
      - FF_APP_KEY
      - FF_APP_ENV=local
      - FF_DB_CONNECTION=sqlite
      - FF_DB_NAME=/var/www/firefly-iii/storage/database/database.sqlite
      - TRUSTED_PROXIES=**
    image: jc5x/firefly-iii
    container_name: firefly
    restart: always
    expose:
      - 80
    volumes:
      - "./firefly-iii/export:/var/www/firefly-iii/storage/export"
      - "./firefly-iii/upload:/var/www/firefly-iii/storage/upload"
      - "./firefly-iii/database:/var/www/firefly-iii/storage/database"
```

@JC5

### DIFF
--- a/.env.docker
+++ b/.env.docker
@@ -19,9 +19,9 @@ TRUSTED_PROXIES=${TRUSTED_PROXIES}
 
 # Database credentials. Make sure the database exists. I recommend a dedicated user for Firefly III
 # If you use SQLite, set connection to `sqlite` and remove the database, username and password settings.
-DB_CONNECTION=mysql
+DB_CONNECTION=${FF_DB_CONNECTION:-mysql}
 DB_HOST=${FF_DB_HOST}
-DB_PORT=3306
+DB_PORT=${FF_DB_PORT:-3306}
 DB_DATABASE=${FF_DB_NAME}
 DB_USERNAME=${FF_DB_USER}
 DB_PASSWORD=${FF_DB_PASSWORD}


JC5:
Please see the discussion here: https://github.com/firefly-iii/firefly-iii/pull/1184

Obbi89:
Well I get your frustration. What I do not understand however is why do you support sqlite in the first place?
I thought it's an oversight that the docker container is bound to mysql only, especially since you are using a database abstraction layer, that supports many different databases.
Same goes for postgresql etc. what if people do have a running psql instance they want to use with ffly3?

So am I getting it right that you do not want to support anything else then mysql?
That is fine, but It isn't clearly communicated in the documentation.

Regarding the separation of concerns. 
You reach that goal by:
a) using a volume for the location of the sqlite database or using a separate service for your database (as you are doing with mysql, could be done with msql or postgresql, too)
b) you strip the apache/nginx out of the base laravel container and use php-fpm to expose an fastcgi socket
c) optionally supply a webserver service that is configured to talk to that container via fastcgi

Then ffly3 would be completely compatible with a bunch of webservers and databases.
You could still tell people you officially do not support anything else than apache2 and mysql and if they have issues in ffly3 due to that you won't help them. Which again is fine, nobody is expecting you to. It's different when we talk about premium support that people pay you for, but as far as I can see you do not offer that in the first place.

JC5:
I'm not sure what else I can add that I haven't said in the other ticket already. I can see postgres happening in the future but so far nobody has actually asked me about it.

There are two kinds of users who use Docker. The one type spins up the compose file and is done. Maybe, just maybe they have a communal mysql instance that handles all DB instances for all their images but those instructions in the docs contained a massive error for Firefly III and nobody noticed for months.

Then there are the more advanced users who (rightfully) want to customize everything. I can cater to both with sensible defaults and I'm willing to. But SQLite support in a docker is just not going to happen. 

Exposing a fast-cgi socket would mean that 99% of Firefly III users get stuck.

Obbi89:
Yeah it wasn't sqlite only. My changes just allow to set the database type manually.
But I think I understand the problem now. 
You just want docker to be idiot proof.

Apart from that you should btw add to the sqlite instructions here
https://firefly-iii.readthedocs.io/en/latest/support/faq.html#i-want-to-use-sqlite
That using sqlite will destroy the user experience and might make imports unusable.
Even in the `self hosted` setup.
(I was experimenting with it for some time, and my guess is during import you are doing a lot of single transactions that need a lock on the database. Enough to make this unusable. The import blocks the database, so that any other request hits the 30 second timeout while it is running.
We tested importing around 800 transactions and it took about 20 minutes. With mysql it was about 3-4 minutes and the system was usable at the same time. That is an issue with file locking being a really inefficient way of synchronization. But that's how sqlite has to do it. There is not much you can change about that behavior without a major rework in the database driver, that isn't part of your code in the first place.)

JC5:
You're right and I shall do so. I will also add a notification that it might break and support will be best-effort. That's only fair :).

Thanks for your feedback so far, I appreciate it! :+1:

PR #1346, opened 2018-04-15
Make database connection configurable for docker

janLo:
This adds the option to configure DB_CONNECTION and DB_PORT via docker
env variables, so the usage of the Image is not limited to mysql.

### DIFF
--- a/.env.docker
+++ b/.env.docker
@@ -22,7 +22,7 @@ LOG_CHANNEL=${LOG_CHANNEL}
 
 # Database credentials. Make sure the database exists. I recommend a dedicated user for Firefly III
 # If you use SQLite, set connection to `sqlite` and remove the database, username and password settings.
-DB_CONNECTION=mysql
+DB_CONNECTION=${FF_DB_CONNECTION-mysql}
 DB_HOST=${FF_DB_HOST}
 DB_PORT=${FF_DB_PORT}
 DB_DATABASE=${FF_DB_NAME}


JC5:
This has been suggested before, but will not be implemented. Please see https://github.com/firefly-iii/firefly-iii/pull/1254 and https://github.com/firefly-iii/firefly-iii/pull/1184 for a discussion about this.

janLo:
Well, I think thats completely the wrong direction. My usecase is not sqlite at all, I just want to use a postgres instance that I have already running. If you want to prevent prevent people from using sqlite, then write it in the docs and/or test that the option contains a valid value on container startup.

Now I'm not able to use the official docker image because of this. Thank you.

JC5:
That last one is something I hadn't even thought about. I'll pick it up!

Most users I have had surrounding Docker so far are newbies. Not implying you are one of course, but these users need guidance most of all. So disabling all of these options makes it easier for me to do support. 

But indeed, this breaks the image for users who _do_ know what they are talking about. A solution is coming. If you know what would be the best way to forcibly disable sqlite, I am all ears.

JC5:
Hi @janLo. I've pushed the change you suggested, together with a big fat warning for users that decide to use "sqlite" as a connection string. I hope this will allow you to use the official docker image again.

The new release should be live end of the month. I apologize for being stubborn.

janLo:
Yes, this works for me. Thank you and sorry for the late reply.

PR #1401, opened 2018-04-29
Select the matching bunq account from the import. Fixes #1398

### LINKED ISSUE #1398: Automaticly choose account when importing from bunq
I am running Firefly III version 4.7.3

#### Description of my issue:

Currently, when import from bunq you need to make a selection on what bunq account matches what account in firefly:
image
I have multiple bunq accounts, and even more none bunq accounts, making it more difficult to use the import.

I would be nice if the importer, based on, I guess iban, select the potentially correct account. This will make the import process much easier to use.

paul999:
Fixes #1398 

Changes in this pull request:

- Make the selectionBox to select the matching IBAN number when importing.

@JC5


JC5:
Good job, merged!

PR #1409, opened 2018-05-03
Adding Indian rupee

shashankmc:
Adding Indian rupee under asian currencies.

Fixes # (if relevant)

Changes in this pull request:
Indian rupee and symbol added.

- 
- 
- 

@JC5


JC5:
Done!

shashankmc:
Thanks man! I only stumbled upon this yesterday as I was looking for self hosted personal manager. When testing on Sandstorm, couldn't find INR and added it. Brilliant job with this BTW! :)

JC5:
Thanks! Did you see you can add them by hand as well? That might save you from having to reset your database.

shashankmc:
Damn. My bad! I must have skipped it in a hurry. I noticed, just now, that is under options. This is my suggestion though - is there any possibility to have a note, in the getting started page, under currency selector indicating to the user that they can add their currency if missing? Would that make sense?

JC5:
Yes it would. I shall open a ticket!

shashankmc:
Cool. :) I'll see if I can work on it.

Issue #1443, opened 2018-05-23
Bunq support in v4.7.4


JC5:
@OGKevin @hippogrief @gldrenthe89 @paul999 @jonathanjuursema @Xeli @IcyPalm @Want100Cookies @FreekBloemhof @japsikado @SpaceK33z 

Hi guys!

I've tagged all of you because you've worked with bunq + Firefly III, or have complained about its implementation. 

Currently live in the `develop`-branch is new working code for the bunq-import routine that works on my machine, as developers tend to say.

I would like to ask you to install a development version of Firefly III and test if it works as expected.

Since I am not sure what your technical expertise is, here are some general instructions.

1. Clone the git repository and check out the develop branch.
2. Copy the `.env.example` file to `.env` and fill in all the details.
3. Check out the `BUNQ_USE_SANDBOX=false` at the bottom. You can switch to the sandbox if you want to.
4. Run the following commands to install all tables and stuff:

```
composer install --no-dev --no-scripts
composer install --no-dev
php artisan migrate:refresh --seed
php artisan passport:install
```

5. Browse to Firefly III. It should self-identify as 4.7.4
6. Try the import!

If you are a Docker fan, [the latest develop build](https://hub.docker.com/r/jc5x/firefly-iii/builds/) should contain the new Bunq code as well but you have to have a [database already](http://firefly-iii.readthedocs.io/en/latest/installation/docker.html).

I am currently writing tests, so be sure to `git pull` regularly.

Thanks in advance for your help!

Cheers,
James

OGKevin:
Will give it a shit later today 👍

OGKevin:
references #1387

japsikado:
Got error when i click the submit button in bunq import section

(1/1) BunqExceptionGenerating a new private key failed. See "http://php.net/manual/en/function.openssl-pkey-new.php" for more info.
--

freshly installed 4.7.4 from develop branch

OGKevin:
@japsikado can you paste the full traceback and on what system are you.

JC5:
Thanks for letting me know! What OS are you using?

japsikado:
I am using windows 10
have installed it on xampp with vhost to public folder
version 4.7.4

`Debug information generated at 2018-05-23 13:59:41 Europe/Amsterdam for Firefly III version **4.7.4**.

| Variable | Content |
| --- | --- |
| FF version | 4.7.4 |
| FF API version | 0.3 |
| App environment | local |
[...]

PR #1445, opened 2018-05-25
Update curl in docker firefly-iii/firefly-iii#1443

OGKevin:
References #1443 (if relevant)

Changes in this pull request:

- Updates curl in docker file.

@JC5


OGKevin:
@JC5 ah never thought of that 😁 thats not a bad idea at all.

PR #1491, opened 2018-06-17
Added support for YunoHost

anmol26s:
Hi 
I have packaged a Firefly  III app for YunoHost. It has install,remove,update,backup,restore script along with multi-instance(can be installed multiple times on different domains on a single server). The app is tested and works well. 
The CL passing will take a day to show up on the app page.
Thanks for coding this app.

Fixes # (if relevant)

Changes in this pull request:

- 
- 
- 

@JC5


JC5:
This is looking pretty good already. Is there some way for me to gain access to the repository?

anmol26s:
Yes 
Yes I have sent you the request.

JC5:
Cool! I will push some things tonight!

Do you know of a way to rename the repository? I would prefer “firefly-iii”, it is more consistent with the name.

anmol26s:
I have given you admin rights. Please don't remove me from the reop. Yes the name can be changed, but I have used it because its easy to type. But if like "firefly-iii" go ahead change it. Till the package app don't break, its good for me.

JC5:
I won't delete you, no worries. Let me explain why I ask it.

* I am strict about "Firefly III" because I hope that it will become a well-known name, over time. If it is Firefly3, I fear that people confuse it with the version (for example Firefly III is now at version 4.7.4).
* I will not delete you from the repo, and I will leave the credits to you in full. You did all the hard work!
* What I want is to have some control in case you decide to stop, or you have no time, or there is a critical bug or something like that. If you are here, and ready to solve any issue, then it is easy. But should you get bored with PHP one day then I want to be able to "step in".

As far as I am concerned, you retain all credits for your efforts. :+1:

anmol26s:
Its so kind and nice of you to step in and take the responsibility of maintaining the package. There are very few developers who even support for packaging these type of repository as they don't find any benefit in it. I do packaging for the apps I use personally and if any end user can use it and cherish the  the great work done by developers like you then there is nothing like that. I don't care much about credits and I don't think end user will ever care about it.

JC5:
I only just realised: could you re-create this pull request on the `develop` branch?

anmol26s:
I have changed it to develop branch.

PR #1592, opened 2018-08-04
Added Swiss Francs to currency seeder according to ISO 4217

claystation:
Changes in this pull request:

- Added Swiss Francs to seeder

@JC5


JC5:
Hey man, thanks for the pull request.

Since users can add currencies themselves under /currencies I am no longer adding currencies to the default set.

PR #1610, opened 2018-08-14
fixed Dockerfile for cURL and cron

### LINKED ISSUE #1464: Firefly III docker image for Raspberry Pi 3
Hi everyone,
thank you for that app it was axactly what I was seaching for.
I tried to install firefly III on a rapberry pi 3 with docker but my container died before working.
Is there anybody who can help me make it work ?

Also I couldn't be able to docker-compose the architecture.

It might not work on rpi3

ErikFontanel:
Fixes #1464 

Changes in this pull request:

- Fixed installation of cURL
Building the image on RPI broke when running curl with a libcurl version mismatch (compare versions with `curl -V`). This mismatch was also present on other platforms, but for some reason didn't break the image builds. Fixed by [loading the new libraries](https://github.com/ErikFontanel/firefly-iii/blob/efd5ceb40533af47db97d9784716544ff6c7d0ce/Dockerfile#L50-L51)

- Added CORES build argument (defaults to 1) to take advantage of multiple cores when building on a RPI to speed up the build process from ~2h to ~45min. Usage: `docker build --build-arg CORES=4 . `.

- Removed some php extensions which are already included by default in the base image

- Added the crontab in a separate file and run it through `entrypoint.sh` and installed rsyslog as @dicastro mentioned.

@JC5
Tested building the image on a 2015 Macbook Pro and Raspberry Pi 3B both running `Docker version 18.06.0-ce, build 0ffa825`. I haven't tested cURL / bunq because I don't have a bunq API key and can't be bothered to create one ;). But with a broken cURL installation, installing composer fails anyway.

### DIFF
--- a/.deploy/docker/crontab
+++ b/.deploy/docker/crontab
@@ -1,0 +1,1 @@
+* * * * * root /artisan schedule:run >> /var/log/cron.log
--- a/.deploy/docker/entrypoint.sh
+++ b/.deploy/docker/entrypoint.sh
@@ -26,3 +26,6 @@ cat .env.docker | envsubst > .env
 composer dump-autoload
 php artisan package:discover
 php artisan firefly:instructions install
+service rsyslog start
+service cron start
+exec apache2-foreground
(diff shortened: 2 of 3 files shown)


ErikFontanel:
That was fast! I noticed an error that occurred due my own merging though, the original ENV variables disappeared. Fixed it now.

PR #1613, opened 2018-08-15
fixed missing ENV variables

ErikFontanel:
Changes in this pull request:

- Readded missing ENV variables for OpenSSL and CURL versions
- Moved add/copy of firefly down to end of Dockerfile to make better use of Docker image layer caching


JC5:
:+1:

Issue #1623, opened 2018-08-19
Transaction success message contain link to transaction


david-me:
Would you mind if I added a feature that adds a link to the transaction in the success message?

Example: After creating a transaction, the "Success! Successfully created new transaction $name" contains a link to the transaction it's talking about (so one can easily go to it)

skuzzle:
I'd appreciate that too. It would also be great if there were an additional checkbox when creating the transaction that, when checked, would redirect to the created transaction upon creation. This would occasionally save some traffic, especially on mobile. I'd mostly use it to create some links to other transactions after creating a new transaction

JC5:
Yes, that certainly is possible. Feel free to build something.

The normal 'success' message is escaped, so simply adding HTML to the string in [firefly.php](https://github.com/firefly-iii/firefly-iii/blob/master/resources/lang/en_US/firefly.php#L783) would not work. However, blindly accepting raw HTML for _all_ [success messages](https://github.com/firefly-iii/firefly-iii/blob/master/resources/views/partials/flashes.twig#L17) is not something you should do. It would mean you would have to verify ALL success messages ever to make sure they do not have weird stuff in them.

An option could be to flash `success_uri` that contains a link. If that variable is set, the success message would be expanded with that particular link. That way you can easily "upgrade" messages and downgrade gracefully of no URI is present.

For the option @skuzzle, that requires some extra thinking so I'm postponing that for now. I don't want to add another checkbox due to clutter. I like the current flow but there's basically three options you have after submitting a form: back to the form itself, back to where you came from or to the new object. And I'm not sure yet how to present the user with that option the most intuitive way.

JC5:
New release is live. Please open a new ticket when the problem persists. Thanks!

PR #1624, opened 2018-08-20
Can have link in success message to the transaction it refers to

david-me:
Fixes # (if relevant)

Changes in this pull request:
Adds option for messages to contain a link (issue #1623), and makes single transactions use that option

@JC5

### DIFF
--- a/app/Http/Controllers/Transaction/SingleController.php
+++ b/app/Http/Controllers/Transaction/SingleController.php
@@ -369,7 +369,8 @@ class SingleController extends Controller
 
         event(new StoredTransactionJournal($journal, $data['piggy_bank_id']));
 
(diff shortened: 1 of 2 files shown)


JC5:
I was thinking more of something like this (check out the `route()` method)

```php
session()->flash('success', (string)trans('firefly.stored_journal', ['description' => $journal->description]));
session()->flash('success_uri', route('transactions.show', [$journal->id]));
```

And then something like:
```html
{% if session_has('success') %}
  {% if session_has('success_uri') %}
    <a href="{{ session('success_uri') }}">
  {% endif %}
  {{ session('success') }}
  {% if session_has('success_uri') %}
    </a>
  {% endif %}
{% endif %}
```

That way, you'll never have unescaped content in the view and you can keep the existing code (less duplication).

What do you think?

david-me:
Yes, that makes sense, dryer

JC5:
Nice. I’ll merge it tomorrow. Thanks for your work! :+1:

david-me:
Thanks for your feedback/help :)

PR #1649, opened 2018-09-03
2FA QR doesn't show up due to CSP error

hamuz:
Error in console:
Refused to load the image 'data:image/png;base64,...' because it violates the following Content Security Policy directive: "img-src 'self'".

Relevant stackoverflow fix:
https://stackoverflow.com/questions/18447970/content-security-policy-data-not-working-for-base64-images-in-chrome-28

Changes in this pull request:
- Adding data scheme to CSP config of image sources

@JC5


JC5:
:+1:

Issue #1538, opened 2018-07-08
Error while ignoring rules on Spectre Import


philthynz:
**Bug description**
I am running Firefly III version 4.7.5

When trying to run a Spectre import, and choosing to NOT apply the rules, there is an import error. I get this via email:

```
Hi there,

Firefly III v4.7.5 ran into an error: Undefined index: apply_rules

The error was of type "ErrorException".

The error occured on/at: Sun, 08 Jul 2018 13:57:53 +0200.

This error occured in file /opt/firefly-iii/app/Support/Import/JobConfiguration/Spectre/ChooseAccountsHandler.php on line 89 with code 0.

The error was encountered by user #1, <myemail>.

The IP address related to this error is: 10.203.31.47
URL is: https://<myserver>/import/job/configuration/hMl8Op43h2qY
User agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Ubuntu Chromium/66.0.3359.181 Chrome/66.0.3359.181 Safari/537.36

The full stacktrace is below. If you think this is a bug in Firefly III, you can forward this message to thegrumpydictator@gmail.com. This can help fix the bug you just encountered.

If you prefer, you can also open a new issue on GitHub.

The full stacktrace is below:

#0 /opt/firefly-iii/app/Support/Import/JobConfiguration/Spectre/ChooseAccountsHandler.php(89): Illuminate\Foundation\Bootstrap\HandleExceptions->handleError(8, 'Undefined index...', '/opt/firefly-ii...', 89, Array)
#1 /opt/firefly-iii/app/Import/JobConfiguration/SpectreJobConfiguration.php(74): FireflyIII\Support\Import\JobConfiguration\Spectre\ChooseAccountsHandler->configureJob(Array)
#2 /opt/firefly-iii/app/Http/Controllers/Import/JobConfigurationController.php(152): FireflyIII\Import\JobConfiguration\SpectreJobConfiguration->configureJob(Array)
#3 [internal function]: FireflyIII\Http\Controllers\Import\JobConfigurationController->post(Object(Illuminate\Http\Request), Object(FireflyIII\Models\ImportJob))
#4 /opt/firefly-iii/vendor/laravel/framework/src/Illuminate/Routing/Controller.php(54): call_user_func_array(Array, Array)
[...]

PR #1707, opened 2018-09-22
Fix bunq import "Undefined index: apply_rules"

mathieupost:
When using the bunq import and unchecking "apply rules" gave an error. Same as https://github.com/firefly-iii/firefly-iii/issues/1538

@JC5


JC5:
Nice find! And thanks for the fix! :tada: I'll have to check if I run into the same problem but I expect I will.

Could you please create a new PR, and recreate your patch on the `develop` branch? Changing the PR itself has given me mixed results.

mathieupost:
Sure!

PR #1708, opened 2018-09-22
Fix bunq import "Undefined index: apply_rules"

mathieupost:
When using the bunq import and unchecking "apply rules" gave an error. Same as #1538

based on develop branch this time ;)

@JC5


JC5:
:+1:

PR #1712, opened 2018-09-23
remove first slash in mass transaction changes

hamuz:
Fixes # (if relevant)

image

Maybe bug (not verified) when mass editing/deleting transactions when Firefly installed in directory.
Couldn't find similar coding pattern in other js files from a quick look (referencing base in uri),
what should be the recommended pattern in this case?

Changes in this pull request:

- remove first slash in mass transaction changes
- 
- 

@JC5

### DIFF
--- a/public/js/ff/transactions/list.js
+++ b/public/js/ff/transactions/list.js
@@ -67,7 +67,7 @@ function goToMassEdit() {
         baseHref = bases[0].href;
     }
 
-    window.location.href = baseHref + '/transactions/mass/edit/' + checkedArray;
+    window.location.href = baseHref + 'transactions/mass/edit/' + checkedArray;
     return false;
 }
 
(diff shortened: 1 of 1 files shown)


JC5:
I think this was my first solution when confronted with the fact that people had Firefly III installed in subdirectories.

Initially these were people who were running Firefly III from /public/ opening their installation to the world. But reverse proxies use directories like /budgeting/ or /firefly/ or /something/ and those I could not scold :wink:.

Mostly I share the URI generated by Firefly III itself in a variable. You can see it in the source of the index page (browser);

```
var accountFrontpageUri = 'https://demo.firefly-iii.org/chart/account/frontpage';
```

It's not particularly pretty but it works. I'm sure that there is a neat solution (I had trouble making Javascript follow the `<base href="">` instruction) but I am not very good at Javascript ;)

PR #1746, opened 2018-09-30
Fix incorrect logging message

ms32035:


JC5:
Nice, I see how that could be very confusing.

PR #1755, opened 2018-10-01
Initial webhook action implementation

hamuz:
Changes in this pull request:
- Initial webhook action implementation
- Tested against https://webhook.site/
- API compatible object-wise

@JC5


hamuz:
Travis failed because of timeout - ran for 50min and terminated.

PR #1780, opened 2018-10-12
Created updater script for virtual or real server

BoGnY:
Changes in this pull request:

- Created update.sh bash script to perform automatic update to a new FireflyIII version (obviously only for virtual or real server)

I have tested it on my VPS.

@JC5

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -1,0 +1,127 @@
+#!/bin/bash
+
+#
+# FireflyIII updater script for virtual or real server
+#
+# Usage:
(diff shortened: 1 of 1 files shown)


JC5:
I like this, but this is not something I want to add to the Firefly III repository itself. This is something to add to a Gist or link from the instructions. It feel it adds a lot of overhead to the repository.

I am trying to move to a repository that only contains the tool. That would exclude stuff like the Sandstorm files, test scripts and other things. And an updater script doesn't really fit I'm afraid.

Issue #1727, opened 2018-09-25
Allow for different backing storage in Firefly


hamuz:
**Description**
Please describe your feature request:

As a Heroku user, I would like Firefly III to be able to save files (import/export/attachments/logs) in different storage locations. This is important because Heroku has ephemeral filesystem.
Read more at: https://help.heroku.com/K1PPS2WM/why-are-my-file-uploads-missing-deleted

I would prefer to use off-site SFTP or FTP (in that order). 

**Solution**
Describe what your feature would add to Firefly III.

Ability to persist uploaded files after dyno restart.

**What are alternatives?**
Please describe what alternatives currently exist.

I've researched a few:
1) No additional packages: Laravel supports the FTP adapter for storage.
in this option we just edit https://github.com/firefly-iii/firefly-iii/blob/d49a13b0916c6c9ae065bacf8471e95c67877edc/config/filesystems.php#L73-L80
to use ftp driver (with its configuration from environment, if such available).
I envision in this case some `if`-s before the `return` , or some ugly ternary inside the return to get this right.

2) Use fallback driver - requires https://github.com/Litipk/flysystem-fallback-adapter
Allows to fallback to to remote storage if not available locally. Not exactly what we want, because writes are only local, and we need to provide a service provider for Laravel (too lazy for that).

3) Use:
* https://github.com/danhunsaker/laravel-flysystem-others
* https://github.com/Litipk/flysystem-fallback-adapter
* https://github.com/thephpleague/flysystem-replicate-adapter
* https://github.com/thephpleague/flysystem-sftp
First of all first package registers with Laravel different service providers (if available, after including in composer.json).
Second of all it provides the 'Mirror' meta-adapter which combines fallback adapter (mentioned before), and replicate adapter to fallback on reads, and also to distribute writes.
https://github.com/danhunsaker/laravel-flysystem-others/blob/43c5e3bf8cbb9bc22a2e228cb644882f0be52f29/config/filesystems.php#L297-L300
Third, and last, this opens up the possibility to easily use wide range of backing storage in the future (like Dropbox, for example).
https://github.com/danhunsaker/laravel-flysystem-others/blob/43c5e3bf8cbb9bc22a2e228cb644882f0be52f29/config/filesystems.php#L40-L44

My recommendation is to use the last option, because it's the easiest (drop packages to composer and edit filesystems.php) and allows many options in the future. I'd be happy to implement this with your blessing.

[...]

PR #1782, opened 2018-10-13
fix local references in upload/export disk. first step for #1727.

hamuz:
Related to #1727 (if relevant)

Changes in this pull request:

- fix local references in upload/export disk

@JC5


JC5:
:+1:

PR #1787, opened 2018-10-14
allow sftp host as backing storage using env

### LINKED ISSUE #1727: Allow for different backing storage in Firefly

hamuz:
Fixes #1727 
Changes in this pull request:

- Added SFTP env variables to enable functionality
- SFTP_HOST and SFTP_FIREFLY_ROOT are required to enable

@JC5


hamuz:
I've typo on line 162 in config/filesystems.php
Should be export_disk, instead of upload_export.
I'd be thankful if you could fix it. With this fix  it should work as intended.

JC5:
Looks good! I like the mirror construction as well.

Instead of `SFTP_FIREFLY_ROOT` however, I would suggest adding `SFTP_EXPORT_PATH` and `SFTP_UPLOAD_PATH`. That gives more flexibility and tells users they need to have those paths available. 

Shall I can merge this and tinker with it?

JC5:
I threw some things around in the config file. Nothing terribly exciting. It should start working once you set the `SFTP_HOST` in your config file. I've "split" the path to the export/upload directory and I've added the private key option. 

I'll expand the other config files as well and then it might already be done! :+1:

JC5:
Also pushed updated config files.

PR #1800, opened 2018-10-17
bills: match attachment route with the icon

hamuz:
Fixes # (if relevant)

Changes in this pull request:

- bills: match attachment route with the icon

@JC5


JC5:
Nice, thanks!

PR #1823, opened 2018-10-26
Using translation of "start balance" for initial balance transaction

### LINKED ISSUE #1815: Little missing translation
**Bug description**
I am running Firefly III version 4.7.7

Just set any languages different from english, and as visible in screenshot, in red circle the translation in missing.. I've searched in Crowdin the text "initial balance" but there are 0 corrispondent string...

firefly

So I think the translation isn't set on php code page 👍

dennisenderink:
Fixes #1815 

Changes in this pull request:

- added translation to FireflyIII\Services\Internal\Support\AccountServiceTrait::storeOpposingAccount()

Maybe a bit simplistic approach, other option is to just use "Initial balance" or a new translation. @JC5

### DIFF
--- a/app/Services/Internal/Support/AccountServiceTrait.php
+++ b/app/Services/Internal/Support/AccountServiceTrait.php
@@ -35,6 +35,7 @@ use FireflyIII\Models\TransactionJournal;
 use FireflyIII\Models\TransactionType;
 use FireflyIII\Services\Internal\Destroy\JournalDestroyService;
 use FireflyIII\User;
+use Illuminate\Support\Facades\Lang;
 use Log;
 use Validator;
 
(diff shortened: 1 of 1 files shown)


JC5:
Nice! A few things I see:

* Could you open a new PR on the `develop` branch?
* You can use `(string)trans('firefly.something')` so there is no need for extra imports. The `(string)` is necessary because `trans()` could theoretically return an array.

Since you're doing translation work with a variable, you could also do something like this:

```php
$name = trans('firefly.something', ['name' => $name]);
```

And the translation string would be: `:name initial balance`.

This gives me the freedom to translate `Startsaldorekening van :name`.

Let me know what you think :+1:

dennisenderink:
I will update this pull request with your feedback. Changed the base to `develop` already so you can reopen it if you want!

JC5:
That has given me a lot of mess in the past. Not sure if GitHub can handle it well. I know they added it as a feature, but my Git history is spaghetti :grin:

I would appreciate a new PR.

dennisenderink:
Alright! Will do!

PR #1929, opened 2018-12-26
Fix fullcalendar.min.js links

a1ex4:
Fixes # (if relevant)
Should be good now 👍 
Changes in this pull request:

- Fixed fullcalendar.min.js link in edit.twig
- Fixed fullcalendar.min.js link in create.twig

@JC5


JC5:
:+1: cheers!

PR #1977, opened 2019-01-07
Added memcached support and zend opcache

BastiaanN:
Changes in this pull request:

- Added libmemcached-dev package to apt-install
- Added opcache to php-extensions to install
- Added pecl installation of memcached-3.1.3 (used explicit version as recommended on https://hub.docker.com/_/php/ )

Memcached can be used by providing the following environment variables:
MEMCACHED_PERSISTENT_ID=[random-string-unique-for-installation]
MEMCACHED_HOST=[adres-of-memcached-server>]
CACHE_DRIVER=memcached

@JC5

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,8 @@ RUN apt-get update && apt-get install -y libpng-dev \
                                             gettext-base \
                                             libldap2-dev \
                                             libpq-dev \
-                                            locales && \
+                                            locales \
+                                            libmemcached-dev && \
                                             apt-get clean && \
                                             rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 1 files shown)


JC5:
In order for this to work, make sure that you update `.env.docker` and change the following rule:

```
CACHE_DRIVER=file
```
To:
```
CACHE_DRIVER=${CACHE_DRIVER}
```

Otherwise, the environment variable will be overwritten by the value in `.env`.

BastiaanN:
By default the PHP library used by Laravel (DotEnv) to read the .env file does not override existing variables already present in the environment, so this seems unnecessary to me.  I verified the application does not start when supplying an invalid value on the docker command-line.

For example, when running the following command (with the docker image build from my branch):
```
docker run -e FF_APP_KEY=S0m3R@nd0mString0f32Ch@rsEx@ct1y -e CACHE_DRIVER=thedriverthatdoesntexist bastiaann/firefly-iii:latest
```

The last error is this one:
`Cache store [thedriverthatdoesntexist] is not defined.`

JC5:
That’s good to know! Thanks.

PR #2048, opened 2019-02-03
Fix delete-account confirmation

nescafe2002:
Steps to reproduce:
- Go to Options, Profile, Delete account
- Enter password
- Click "DELETE your account"
- Confirmation dialog appears: "Are you sure? You cannot undo this." (OK/Cancel)
- Click **Cancel**

Expected behavior:
- Account is not deleted

What happens instead:
- Account is deleted

Changes in this pull request:
- Do not delete account if user cancels confirmation dialog

@JC5


JC5:
Ah nice find. Thanks! Merged.

Issue #1957, opened 2019-01-01
Split a piggy bank into multiple asset accounts


lastlink:
**Description**
Please describe your feature request:

- Difficult to do some piggy banks, b/c might have cash on hand as well amounts in a savings assets set aside for same piggy bank

**Solution**
Should be able to add a piggy bank to multiple assets and track amounts in each allocated to the piggy bank

**What are alternatives?**
I don't know if #18 is part of this.

JC5:
That can be done but it requires a pretty large database change. I'll think about the best way to do this.

jasonbone75:
I'll add a vote to this. My use case is setting aside tax and NI each month when self-employed using regular savers attached to current accounts - you have a limit on deposits so have to spread across multiple accounts but still want to track progress towards the 6 monthly payment deadlines.

JC5:
https://github.com/firefly-iii/firefly-iii/projects/5#card-19974876

lastlink:
What is the motivation behind closing the issues that are linked in the project cards? why not add the issues directly into the project kanban or convert the cards into issues to track easier. we can't add comments to project cards that aren't issues.

JC5:
It cleans up the list of issues. I can refer people to the roadmap using the issue template.

PR #2123, opened 2019-02-25
F/split piggy

lastlink:
Fixes issue #1957 

Changes in this pull request:

- adds transfer, account_id, from_account_id to piggy_bank_events table
  - also auto sets account_id to piggy bank default, db rollback works great (mysql only one tested)
- allows piggy banks to be used by multiple accounts
- didn't setup dropdown, but when adding or subtracting from piggy bank will use default account set to track
- shows total for each bank account in piggy bank display page (only for accounts that have had amounts added or subtracted, if have only been transferred from as source then the aggregation sql query won't include them)
  - sql aggregate query has been tested on **mysql** several times and calculates sum of each account for a piggy bank correctly, may need to be tweaked for postgres & sqlite.
- transfers only no longer change piggy bank current amount
- deposits and withdraws should now be supported, just need to update expandform to show on them (I could not figure this out)
- didn't yet do a display of piggy bank amounts in accounts/show next TODO

new piggy bank view

@JC5

### DIFF
--- a/app/Factory/PiggyBankEventFactory.php
+++ b/app/Factory/PiggyBankEventFactory.php
@@ -19,7 +19,7 @@
  * along with Firefly III. If not, see <http://www.gnu.org/licenses/>.
  */
 /** @noinspection MultipleReturnStatementsInspection */
-declare(strict_types=1);
+declare (strict_types = 1);
 
 namespace FireflyIII\Factory;
 
@@ -84,8 +84,10 @@ class PiggyBankEventFactory
 
             return null;
         }
-
-        $piggyRepos->addAmountToRepetition($repetition, $amount);
+        // only add if not Transfer
+        if (!$journal->isTransfer()) {
+            $piggyRepos->addAmountToRepetition($repetition, $amount);
+        }
         $event = $piggyRepos->createEventWithJournal($piggyBank, $amount, $journal);
         Log::debug(sprintf('Created piggy bank event #%d', $event->id));
 
--- a/app/Http/Controllers/Chart/PiggyBankController.php
+++ b/app/Http/Controllers/Chart/PiggyBankController.php
@@ -52,6 +52,32 @@ class PiggyBankController extends Controller
         $this->generator = app(GeneratorInterface::class);
     }
 
+    public function currencies($curr): JsonResponse
+    {
+        $curl = curl_init();
+
+        curl_setopt_array($curl, array(
+            CURLOPT_URL => "https://api.exchangeratesapi.io/latest?base=".$curr,
+            CURLOPT_RETURNTRANSFER => true,
+            CURLOPT_ENCODING => "",
+            CURLOPT_MAXREDIRS => 10,
+            CURLOPT_TIMEOUT => 30,
+            CURLOPT_HTTP_VERSION => CURL_HTTP_VERSION_1_1,
+            CURLOPT_CUSTOMREQUEST => "GET",
+        ));
+
+        $response = curl_exec($curl);
+        $err = curl_error($curl);
+
+        curl_close($curl);
+
+        if ($err) {
+            return response()->json("error");
+        } else {
+            return response()->json($response);
+        }
+    }
+
     /**
      * Shows the piggy bank history.
      *
--- a/app/Models/PiggyBank.php
+++ b/app/Models/PiggyBank.php
@@ -85,8 +85,8 @@ class PiggyBank extends Model
         if (auth()->check()) {
             $piggyBankId = (int)$value;
             $piggyBank   = self::where('piggy_banks.id', $piggyBankId)
-                                ->leftJoin('accounts', 'accounts.id', '=', 'piggy_banks.account_id')
-                                ->where('accounts.user_id', auth()->user()->id)->first(['piggy_banks.*']);
+                ->leftJoin('accounts', 'accounts.id', '=', 'piggy_banks.account_id')
+                ->where('accounts.user_id', auth()->user()->id)->first(['piggy_banks.*']);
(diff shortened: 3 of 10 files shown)


JC5:
Hey man, thanks for this. I like what you've built, but it's not viable for inclusion in Firefly III. Please give me a heads-up before you build something big like this.

If you want to create multi-account piggy banks you need to think of a few constraints first.

* Accounts can have different currencies.
* You need a single source of truth.
* The API must also support it.
* You need to be able to migrate towards it.
* It's an excellent moment to clean up some old code.
* I want to get rid of piggy bank repetitions. 
* Piggy bank events are informational only

So first things first. Let's assume a user wants a piggy bank and save for a couch on two different accounts. The first thing you need is a "piggy bank group" that holds meta information like the title, description, total amount and currency. Also target date etc. Then, for each account that's part of a group you need a piggy bank. Each has a current amount, a link to the group and a related account. It doesn't hold the currency, that's done by the account it's linked to.

When the user creates a piggy, they actually create a piggy bank group and X piggy banks, depending on which accounts the user selects during the creation process. There is a constraint here that all accounts must have the same currency but it's not a hard constraint: the user can change the currency later and the piggy bank group will just have to deal with it. Likewise when the user edits a piggy bank group the piggy banks are created and deleted if they change the accounts. This should all be transparent to the user (invisible). But it must be easy to use so help texts are required and multi-select things used.

Overview of all piggy banks is simply a list of all the piggy bank groups and the total amount compared to the piggy bank. Must take care that a piggy bank can't overflow. And how will you handle multiple currencies? You can't do a currency conversion because you can't rely on a working internet connection to do the conversion for you. Will you remove the account from the piggy bank group? Because that includes magic that's not very user friendly.

Piggy banks events are nice but only for informational purposes. You can't rely on them to keep track of the piggy bank accurately. Each piggy bank in a group is the single source of truth for the amount in the piggy bank. This makes calculation of amounts a lot easier too. The only reason you can link transfers to piggy banks is because it's easier for the user but I won't extend it to the other transaction types.

Then, we must also change the API to accomodate this.

So in it's current form this is not something I can include, because it's different from the way I'd like to build it. I'll try to set something up in `develop` this week and share it.

lastlink:
The current piggy banks also have some issues, account balances can be changed and not reflect what's currently in the piggy banks. An account could be deleted and the piggy bank will be inaccurate. Could piggy banks not be lightly enforced such as if a user deletes an account they may have to delete the piggy bank to fix it. For the most part piggy banks are "informational" to help a user organize their finances such as something you might lightly put together in excel.

Also I think piggy banks should only support one currency so if a bank account is in a different currency it won't be permitted. Maybe set a currency for the piggy bank and only allow accounts of that currency. It doesn't make sense to split a piggy bank across accounts of different currencies.

I don't understand what piggy bank repetitions are, are these created from bills, rules, or recurring transactions? Or are these related from transfers. I'd propose that a piggy bank could be used on any transaction: deposit, withdrawal,  transfer. Only transfers wouldn't change value, b/c they haven't left your accounts.

But thanks for clarifying constraints needed for this to be done. I see this as very useful being able to track savings goals across multiple accounts and put some work towards it b/c of it's importance.

I'm also willing to put in a pull request on the docs piggies.html when we're done with this.

Thank you for your help!

JC5:
Like you say, piggy banks are supplemental, so managing accounts should not interfere with the piggy banks. If that means that there's too little money to fill the piggies, it's up to the user to correct it.

But the point is, it could be that piggy banks inadvertently support multiple currencies: if you create it first and then edit the account, it could happen. So it's something to consider.

Piggy bank repetitions are a leftover from an old feature where piggy banks would auto repeat themselves (for yearly savings, for example). But it was terribly complex and not very user friendly.

lastlink:
No where in your application as far as I'm aware do you do currency conversion. Piggy banks simply don't make sense to do it unless you convert it to a base currency. Your **net worth** and other graphs show the currencies as different totals. Piggy banks b/c they have a goal don't make sense with multiple currencies unless converting the rates.

If you wanted to supported multi currency the piggy bank current value should be calculated and not stored in the database. 
You could still track event with the piggy bank and do totals for each account in the accounts currency. The piggy bank would have a default currency & you could use js or php to call something like https://fixer.io/signup/free rest api  (or have the user enter the conversions into the database for the default currency) and convert all the currencies to the piggy bank currency and calculate the current total.

Or with one currency when doing a transfer only show piggy banks in that currency and only allow selection of accounts (when editing a piggy bank) of the same piggy bank currency.

What would be your preference on this?

lastlink:
Last thoughts,
I'm a big fan of minimal database changes. My last change added only 3 table columns. I think going the multi-currency route we could use javascript (rather than php to calculate post page render) from https://exchangeratesapi.io/ to calculate the conversion for each account to the currency of the current selected account for a piggy bank and use this calculated total. Of course the currency must be available in the rest api. I could make this change by updating the https://github.com/firefly-iii/firefly-iii/blob/master/public/v1/js/ff/piggy-banks/show.js and adding appropriate id's into the show.twig file.

What are your thoughts on me going this route?

The rest api for ff-III will need to expose the new fields in piggybankevents and maybe the piggy bank accounts aggregation calculation.

The reason for piggy bank events for other transaction types is if a user wants to use the piggy bank their couch purchase and empty it out.

JC5:
Let’s leave it like it is right now. I’ll build something later. This isn’t feeling right, it feels cobbled together and quick & dirty.

lastlink:
Ok I added javascript to calculate the currency amounts for each and do new totals from this. 
currencycalculation

JC5:
Like I said, I'll try to find the time to build something. Don't pursue this for the time being.

Issue #2125, opened 2019-02-25
upgrade failure in 4.7.9 to 4.7.14


GaneshKandu:
**Bug description**
I am getting Error while upgrading from **4.7.9** to **4.7.14**

```
FireflyIII
installation and upgrade

Please wait...
An unknown error prevented Firefly III from executing the upgrade commands. Sorry. No currency found with code "eyJpdiI6IksyWlZzUk9kMFRlUFwvdFlYdTRrem5BPT0iLCJ2YWx1ZSI6IkNSWnBOUkpVXC9HcjJJeXJzYzg1RWxRPT0iLCJtYWMiOiI2NTk3MDY4ZmU5YTg4YjY3YmIzNmU3NDg2Y2MxNGIxM2E5MTY3MTZmM2ZlMGQ2Y2I3M2Y1MDM5MGVhNTUzZjA5In0="

Please read the official documentation about this, and upgrade by hand.

```

**Steps to reproduce**
upgrade from 4.7.9 to 4.7.14 using web

**Extra info**
System Environment

.| Version
--- | ---
PHP | 7.2.8
MySQL | 5.6.42
Server API  | Apache 2.0

**Screenshotinfo**

![screenshot-webuzo2 web-2019 02 25-12-20-12](https://user-images.githubusercontent.com/12833130/53319850-c7b43900-38f9-11e9-9999-2414e808a892.png)

JC5:
Mmmm, that looks like a bug. I don't think I can fix it from here. You'll have to wait for the next release (upcoming weekend) so I can build a fix and test it. Sorry about that.

JC5:
I've pushed a fix to the develop branch. Are you capable of trying it out?

lastlink:
This may be related, but I had an issue when updating beyond 4.7.6 b/c I had done a rollback from 4.7.7. since I was still on 7.1 at the time. It was because there wasn't actually a `down() `implementation of the 477 migration.

https://github.com/firefly-iii/firefly-iii/blob/master/database/migrations/2018_09_05_195147_changes_for_v477.php

GaneshKandu:
Hi,

Just Checked with ```firefly-iii-develop``` zip but still getting same issue in upgrade

```

FireflyIII
installation and upgrade

Please wait...
[...]

PR #2130, opened 2019-02-26
add missing rollback for 477 migration

lastlink:
Fixes issue #2125 maybe

Changes in this pull request:

- rollback implementation for migration 477

@JC5


JC5:
It shouldn't be necessary to add the rollback because the associated table (transaction_currencies) is always dropped after this table is dropped. I'll do some testing first.

JC5:
Wait, I forgot you still need to be able to re-apply it 😅 . Merged!

PR #2174, opened 2019-03-14
F/split piggy

lastlink:
Fixes issue #1957 

an update on pull #2123. I know this will probably be rejected, until other priorities are done. I'm now using this b/c being able to organize piggy banks across multiple accounts is super useful for some things that I'm doing and I thought I'd show my changes.

You can rollback the db migrations safely before a new version update by doing 
`php artisan migrate:rollback --step=1` and just note how your accounts were organized

#### ignore these if you don't wish to have the multi currency account support
```
resources/views/v1/transactions/single/create.twig
public/v1/js/ff/piggy-banks/show.js
app/Http/Controllers/Chart/PiggyBankController.php
```

Changes in this pull request:

- allows piggy banks to support multiple accounts and currencies.
- adds transfer, account_id, from_account_id to piggy_bank_events table
  - also auto sets account_id to piggy bank default, db rollback works great (mysql only one tested)
- allows piggy banks to be used by multiple accounts
- didn't setup dropdown, but when adding or subtracting from piggy bank will use default account set to track
- shows total for each bank account in piggy bank display page (only for accounts that have had amounts added or subtracted, if have only been transferred from as source then the aggregation sql query won't include them)
  - sql aggregate query has been tested on **mysql** several times and calculates sum of each account for a piggy bank correctly, may need to be tweaked for postgres & sqlite.
- transfers only no longer change piggy bank current amount
- deposits and withdraws should now be supported, just need to update expandform to show on them (I could not figure this out)
-also displays  piggy bank amounts in accounts/show
- piggy bank total in the piggy-banks/show/ is calculated using currencies for each account

Account status amounts on page /piggy-banks still needs to be updated

@JC5

### DIFF
--- a/app/Console/Commands/VerifyDatabase.php
+++ b/app/Console/Commands/VerifyDatabase.php
@@ -348,7 +348,8 @@ class VerifyDatabase extends Command
                 }
 
                 $type = $journal->transactionType->type;
-                if (TransactionType::TRANSFER !== $type) {
+                // should be one of the 3 types
+                if (TransactionType::TRANSFER !== $type && TransactionType::DEPOSIT !== $type && TransactionType::WITHDRAWAL !== $type) {
                     $event->transaction_journal_id = null;
                     $event->save();
                     $this->line(sprintf('Piggy bank #%d was referenced by an invalid event. This has been fixed.', $event->piggy_bank_id));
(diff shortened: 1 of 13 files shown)


JC5:
Good work! Alas, as you say, this will not be included.

viharm:
@JC5, Although I am not particularly keen on multi-currency piggy banks, I am desperate to have a multi-account piggy bank. I am confident that this is a very real scenario and people may have savings different accounts all for the same target. One of the real constraints is that high interest savings accounts have a limit on the deposit each month or each year, which means even if you have the money, you can't put it all in the same account.

I didn't want to create a separate issue, as you have already mentioned that you have "put a pin" on this, but thought it would be good to let you know that there is appetite for this.

Thanks for the great work!

PR #2209, opened 2019-04-04
CSV file fix for Belfius

SanderKleykens:
This pull request adds a CSV file fix option for Belfius Bank (Belgium). 

### Problem
Belfius CSV files have, among other columns, a column containing the transaction description ("mededeling") and a "transaction" column, containing a short summary of the transaction (which also includes, but is not limited to, the transaction description or "mededeling").
For outgoing recurring transactions, the "mededeling" column contains the same short summary of the transaction as the "transaction" column.
This causes problems when the outgoing recurring transaction is a recurring transaction to another account that gets imported as well. The transaction descriptions can't be matched since the "mededeling" column for the outgoing transaction contains more than just the transaction description, unlike the receiving account's transaction. This results in duplicate transactions being imported.

### Implementation
In the fix implemented by this pull request, the "transaction" column containing a short summary of the transaction remains as is. However, the transaction description ("mededeling") column of recurring transactions gets adjusted to only contain the transaction description ("mededeling"), so it matches the description found for the transaction at the receiving end.

### DIFF
--- a/app/Import/Specifics/Belfius.php
+++ b/app/Import/Specifics/Belfius.php
@@ -1,0 +1,94 @@
+<?php
+/**
+ * Belfius.php
+ * Copyright (c) 2019 Sander Kleykens <sander@kleykens.com>
+ *
+ * This file is part of Firefly III.
+ *
+ * Firefly III is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Firefly III is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


JC5:
Thanks. I’ll check it out this weekend. You can drop all the translations from your PR, I use Crowdin to manage those.

SanderKleykens:
Thanks and done :)

JC5:
The code is looking well, nice job! 

I was a bit to fast with my earlier comment. Could add the en_US strings to your PR again? No need to add the others. That will properly trigger Crowdin. Sorry for the confusion 😅

SanderKleykens:
No problem, @JC5! Fixed it :)

PR #2281, opened 2019-05-22
Update firefly.php

lucianor:
Changes in this pull request:
- Updated wrong translation. 'Left' is translated in Portuguese to the opposite of 'right' ("Esquerda"), when it should be translated to mean remaining ("Restante")


JC5:
Thanks for pushing the update. But translations are handled through [Crowdin](https://crowdin.com/project/firefly-iii), and if you register there you can fix the translation.

I'll automatically include it in the next release :+1:

PR #2303, opened 2019-06-08
Correct ratesapi URL

JurajMlich:
Changes in this pull request:

- Correct the endpoint for ratesapi. The current one returns 500, the new one works correctly.

@JC5


JC5:
Nice, thanks.

PR #2318, opened 2019-06-23
CSV file fix for ING Belgium

SanderKleykens:
This pull request adds a CSV file fix option for ING Bank (Belgium).
ING Belgium stores the opposing account name, IBAN and the transaction's description together in 1 column in their CSV exports instead of multiple columns that can be mapped to transaction properties in Firefly.
This extracts that information into separate columns.


JC5:
:+1:

Issue #2353, opened 2019-07-29
Repeated valid transactions should be allowed on import


viraptor:
**Bug description**
I am running Firefly III version 4.7.17.3, and my problem is:

When importing from CSV file I'm running into a situation where there are multiple valid entries with the same date/description/amount. For example:
```
24/07/2019,"-1.49",APPLE ITUNES STORE        SYDNEY
24/07/2019,"-1.49",APPLE ITUNES STORE        SYDNEY
```
or multiple tram rides which are separate lines:
```
24/05/2019,"-6.76",TRANSPORTFORNSW TAP       SYDNEY
24/05/2019,"-6.76",TRANSPORTFORNSW TAP       SYDNEY
```
Currently firefly doesn't like importing the repeated entries, but this does happen quite a lot for me. (12 over 3 months, mostly from taking public transport)

This was previously mentioned in https://github.com/firefly-iii/firefly-iii/issues/1037

**Steps to reproduce**
Import lines mentioned above.

**Expected behavior**
I see that this is an interesting issue and not everyone may have the same expectations here.
I'd like to propose a resolution which I used in previous previously in my custom statement processing:

- Assume that when importing, the file will have a full day included (there won't be missing entries for a given date if another transaction is available there)
- Merge the number of entries. For example if the imported statement contains 3 entries, and the database does not, add 3 entries. If the import contains 3 entries and the database contains 2, add 1 entry as new.
- Possibly warn if the database contains more matches than the import

This also allows easy import of CSVs where the bank makes old entries suddenly appear, for example when pending transactions don't resolve all in the same time. (Yes, that kind of shows two use cases as one really)

Maybe this could be possible to implement behind some extra import option ("allow duplicates and merge"?)

I'm happy to chat about alternative approaches.

JC5:
If Firefly III can't distinguish individual lines they'll be treated as a duplicate, and they won't be imported. This is way more useful for a lot of people than the assumption there is no overlap between imports. The logic to add 2 out of 3 entries is unsustainable and very complex. Your last point could be an option, but not in the current version.

You should approach this from the other end. The lines should be unique. CSV exports should contain at least transaction ID's or other unique values.

viraptor:
Unfortunately the available feeds are really broken :-( My banks don't provide any csv id fields, and even their ofx export has duplicates in the "unique ID" field.

basilgohar:
[...]

PR #2356, opened 2019-07-30
AppendHash specific for handling duplicate CSV entries

viraptor:
Fixes issue #2353

Changes in this pull request:

- Add an AppendHash specific
- It appends an artificial hash column to the CSV imports, which can be used as External ID
- It allows duplicate transactions from a single import, but will also prevent re-importing the same entries from multiple runs (each time hashes are based only on the row contents + duplicate counter)

(This is in WIP stage, I'll sort out the documentation and strings if you think the idea itself is acceptable)

@JC5


viraptor:
@JC5 Is this approach ok? Should I polish the PR for merging?

JC5:
I'm going to postpone this until after the next release, so stay tuned. I only noticed the copyright statement isn't yet in your name.

viraptor:
Step away, bot

JC5:
I'll mark it as a bug. That keeps the bot away. I might take the code into a new project as per the roadmap though.

JC5:
The code has been merged with the new CSV importer here:

https://github.com/JC5/firefly-iii-csv-importer

I'm going to let this one be closed by the bot, and pick up these improvements in the new CSV import routine. For more info, see also: https://www.patreon.com/posts/30012174

PR #2368, opened 2019-08-02
Allow a custom redirect scheme (using redirect)

Sotam:
This Apache config change will add support for a custom (OAuth2) redirect URL scheme. This can be used for e.g. mobile apps.

So, in Firefly OAuth config you can use `https://{your-domain}.com/oauth-callback` as a redirect/callback URL and in an (mobile) app register the custom scheme `org.firefly-iii`.

Please let me know what you think of this, it would help me with my iOS development.

@JC5

### DIFF
--- a/.deploy/docker/apache-firefly.conf
+++ b/.deploy/docker/apache-firefly.conf
@@ -11,6 +11,8 @@
 
         ErrorLog ${APACHE_LOG_DIR}/error.log
         CustomLog ${APACHE_LOG_DIR}/access.log combined
+        
+        Redirect permanent /oauth-callback org.firefly-iii://oauth-callback
 
     <Directory /var/www/firefly-iii/public>
         Options -Indexes +FollowSymLinks


JC5:
Wouldn't that change the redirect for everything?

Sotam:
Only for the https://{firefly-domain}/oauth-callback. Everyone can call that url and the redirect will happen. The best solution would be support for custom OAuth redirect scheme, but that’s a Laravel Passport issue and not really fixible

Sotam:
Although this change would be the easiest change, I've looked into other possible solutions. I could ditch the iOS specific authentication method and use a (custom) webviewer. Later today I'll try a POC and hopefully it works perfectly, so we can close this PR :)

JC5:
I just wanted to say that I'm not going to implement a fix that breaks the redirect for everybody else.

Sotam:
I understand completely and somehow the POC is a better and perhaps safer implementation (like I tried to explain in the e-mail). I figured the redirect isn't 100% accurate, sometimes the redirect never happens. So I'll close this PR.

JC5:
I saw that, haven't had the time to reply yet. Thanks :)

PR #2377, opened 2019-08-08
Uses of APP_NAME in views

danielmarcoto:
Changes in this pull request:

- Adds the use of environment variable APP_NAME throughout the views;
- Changes the controller that set the title to use APP_NAME;

@JC5


JC5:
Interesting project.

PR #2408, opened 2019-08-12
Update firefly.css

al-sh-ac-san:
Fixes issue # (if relevant)

Changes in this pull request:

- 
- 
- 

@JC5


JC5:
What's unfriendly about the font?

al-sh-ac-san:
Sorry! I didn't mean to create a PR for that as it is quite stupud fix. Did that by mistake. Having own fork for that. It is an issue with 'Source Sans Pro' font which is rendered incorrectly in safari on mac and iphone with cyrillic font.

PR #2429, opened 2019-08-15
Fix typo in BudgetRepository (emd_date)

GeoffreyFrogeye:
Changes in this pull request:

- Fixed typo in BudgetRepository which could trigger an SQL error during a GET `/api/v1/available_budgets` with the `end` parameter set. 

@JC5


JC5:
Nice, merged.

PR #2430, opened 2019-08-15
Adding POSTGRES_DB env variable

seanmdalton:
Fixes issue # (if relevant)

Changes in this pull request:

- example docker-compose.yml updated to create the default database as specified in .env.example

@JC5

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,6 +26,7 @@ services:
     environment:
       - POSTGRES_PASSWORD=secret_firefly_password
       - POSTGRES_USER=firefly
+      - POSTGRES_DB=homestead
     networks: 
       - firefly_iii_net
     volumes: 


JC5:
Nice, but I think less is more. Why not change the `DB_DATABASE` value to `firefly` in `.env.example`? That would have the same effect.

seanmdalton:
@JC5 thanks for the quick reply! The default database would still need to be defined/created in the compose file, right? When I pulled the compose and env file, the app container wouldn't start because the database defined in the .env.example didn't exist. I figured adding this made it a bit easier to get the example started. 

I would be for updating both the compose and env files to specify the firefly database. Let me know and I can update the PR.

JC5:
Well because the image defaults to the username as the database name I’d like to avoid that. The composer file aims at less experienced users so the less settings they see the less they can break.

Issue #2441, opened 2019-08-18
API: Currency symbol character limit is not consistent


GeoffreyFrogeye:
**Bug description**
I am running Firefly III version 7b813065da7eded19a70be2806dfcda4e3998647, and my problem is:

The symbol limit on the GUI is 8 characters. However, in the API:

- If you POST a new currency whose symbol is more than 5 characters, you'll get [the error: `The symbol must be between 1 and 5 characters.`](https://paste.frogeye.fr/?0f7ab15355c47286#PPPFXSpQDeksEKYp/2KOo8GEN1POhL6blKrBKzCHn1s=).
- If you PUT an existing currency whose symbol is less than 12 characters, it'll get registered.
- If you PUT an existing currency whose symbol is 12 characters or more, you'll get [a nice SQL error saying that the column is too tight](https://paste.frogeye.fr/?c6babdf5310cf899#0KVt05ZNm+zqSXzwCS+8HaGDIrHxT+/hlEyf6oGaDYU=).

Min is fine though, I mean it will be caught as `The symbol field is required.`.

(mind if I drop the usual debug infos? It doesn't seem relevant, but if you need them I'll be happy to provide them)

JC5:
I merged the PR, let me know if it solved all the issues.

PR #2442, opened 2019-08-18
Fix consistency in symbol character limit

GeoffreyFrogeye:
Fixes issue #2441 

Changes in this pull request:

- Made the character limit for currencies consistent in the API with the GUI 

I swear I looked into the code to see if this was a fix I could do. I should look harder before filling an issue next time :sweat_smile:

However I noticed the rules for `decimal_places`, `enabled` and `default` are not there. Is this intentional (I don't know, some kind of inheritance maybe)?

@JC5


JC5:
I'll have to dive into it later. Thx for this

JC5:
:+1:

JC5:
@GeoffreyFrogeye and yes, the other rules are inherited.

Issue #2443, opened 2019-08-18
API: Server exception on PUT /api/v1/available_budgets/{id} 


GeoffreyFrogeye:
**Bug description**
I am running Firefly III version 7b813065da7eded19a70be2806dfcda4e3998647, and my problem is:

[Editing an existing available budget with the API](https://api-docs.firefly-iii.org/#/available_budgets/updateAvailableBudget) produces a 500 error with the following message:

```
Undefined index: transaction_currency_id
```

[Full JSON output with trace](https://paste.frogeye.fr/?5c30f92bc93f48bf#7sPzHoALxt9S1PdLt/yZW0K9G9UmFdUeWFHvL6ZxJSk=)

**Debug info**

Debug information generated at 2019-08-18 14:44:28 Europe/Paris for Firefly III version **4.8.0.2**.

| Variable | Content |
| --- | --- |
| FF version | 4.8.0.2 |
| FF API version | 0.10.0 |
| App environment | local |
| App debug mode | true |
| App cache driver | file |
| App logging | , daily |
| PHP version | 7.3.8 |
| Display errors | Off |
| Session start | 2019-08-01 00:00:00 |
| Session end | 2019-08-31 23:59:59 |
| Session first | 2015-05-27 00:00:00 |
| Error reporting | ALL errors |
| Host | Linux |
| Interface | fpm-fcgi |
| UserID | 1 |
| Attempt at "en" | false |
| Attempt at "English" | false |
| Attempt at "en_US.utf8" | 'en_US.utf8' |
| Attempt at "en_US.UTF-8" | 'en_US.UTF-8' |
| DB drivers | mysql, pgsql, sqlite |
| Current driver | pgsql |
| Login provider |   |
| Storage disks | local-upload  |
| Using Sandstorm? | no |
| Is Sandstorm (.env) | false |
| Is Docker (.env) | false |
| bunq uses sandbox | false |
| Trusted proxies (.env) |  |
| User agent | Mozilla/5.0 (X11; Linux x86_64; rv:68.0) Gecko/20100101 Firefox/68.0 |
[...]

PR #2444, opened 2019-08-18
Fix wrong indexes in BudgetRepository

GeoffreyFrogeye:
Fixes issue #2443

Changes in this pull request:

- Updated the indexes in BudgetRepository to match the fields of the API

I am **really** not sure this is a correct solution. While the POST call now works, I did not expected the GUI to work too, as  I believed they would share the same code path.

But to my surprise this doesn't break the GUI so I'm submitting this in case this is correct.

@JC5


JC5:
Not all UI code uses the API yet, although this will be happening in the near future. I’ll have to find time to dive into your PR’s later. Thanks again for submitting these fixes. 

For the next release this will push the API to 0.10.1. I know it isn’t exactly semver but ok.

JC5:
:+1:

PR #2471, opened 2019-08-23
Add date attribute to the list row

hamuz:
Used to be there previously (at least at 4.7.8). Used in my automation.
Similar to other views: https://github.com/firefly-iii/firefly-iii/search?q=data-date&unscoped_q=data-date

Fixes issue # (if relevant)

Changes in this pull request:

- 
- 
- 

@JC5


JC5:
It was removed, the ordering doesn't work yet.

hamuz:
What do you mean? Which transaction would show up first?

JC5:
It was used to do the drag-and-drop of transactions on the /accounts/show/x page, which I haven't reimplemented yet.

Issue #1869, opened 2018-11-14
Is APP_URL enough to address the firefly-iii in a different folder?


Odyno:
Hello, 
I'm working with the master version firefly-iii but rebuilt on my raspberry (you can see my branch here https://github.com/Odyno/firefly-iii/tree/rpi-docker  and the docker image here https://hub.docker.com/r/odyno/rpi-firefly-iii/ ).

I'm running the docker image in a compose like that (some variable is removed ;)
```
networks:
  firefly_iii_net:
    driver: bridge
services:
  firefly_iii_app:
    environment:
      - APP_DEBUG=false
      - APP_URL=/finance/
      - TRUSTED_PROXIES=**
      - FF_DB_HOST=...
      - FF_DB_NAME=....
      - FF_DB_USER=...
      - FF_DB_PASSWORD=....
      - FF_APP_KEY=.....
      - FF_APP_ENV=local
      - TZ=Europe/Rome
      - SITE_OWNER=.....
      - MAIL_DRIVER=....
      - MAIL_HOST=....
      - MAIL_PORT=...
      - MAIL_FROM=...
      - MAIL_USERNAME=....
      - MAIL_PASSWORD=...
      - MAIL_ENCRYPTION=...
      - MAPBOX_API_KEY=...
      - FIXER_API_KEY=....

    image: odyno/rpi-firefly-iii:0.0.1
    restart: always
    links:
      - firefly_iii_db
    networks:
      - firefly_iii_net
    ports:
      - "9080:80"
    volumes:
      - "/media/srv/firefly_iii_workspace/export:/var/www/firefly-iii/storage/export"
      - "/media/srv/firefly_iii_workspace/upload:/var/www/firefly-iii/storage/upload"
  firefly_iii_db:
    environment:
      - MYSQL_DATABASE=....
      - MYSQL_USER=...
      - MYSQL_PASSWORD=....
      - MYSQL_ROOT_PASSWORD=....
    image: yobasystems/alpine-mariadb:arm32v7
    restart: always
    networks:
      - firefly_iii_net
    volumes:
      - "/media/srv/firefly_iii/db:/var/lib/mysql"
version: "3.2"
```

and finally I'm exposing the service via Nginx proxy_pass behind the /finance/ route:  

```
  location /finance/ {

        proxy_pass              http://127.0.0.1:9080/;
        proxy_redirect          http://127.0.0.1:9080/ /finance/;
        proxy_pass_header       Server;
[...]

PR #2514, opened 2019-08-29
Allow override of base URL

TimCinel:
Without this reverse proxying outside of the root of a domain requires unreliable workarounds (rewriting HTTP header and even content).

Fixes issue #1869 (and possibly others)

Changes in this pull request:

- Introduce environment variable configuration to optionally force root URL so correct routes are generated when the application is not at the root of a domain (e.g. example.com/firefly)

Without this, all routes are broken when the site is not running at the root path because the routes are produced as absolute URIs. In this situation, most or all redirects are broken and most or all asset references (CSS, JS, images) are broken.

@JC5

### DIFF
--- a/config/firefly.php
+++ b/config/firefly.php
@@ -140,6 +140,7 @@ return [
     'fixer_api_key'                => env('FIXER_API_KEY', ''),
     'mapbox_api_key'               => env('MAPBOX_API_KEY', ''),
     'trusted_proxies'              => env('TRUSTED_PROXIES', ''),
+    'baseurl_override'             => env('BASEURL_OVERRIDE', ''),
     'search_result_limit'          => env('SEARCH_RESULT_LIMIT', 50),
     'send_report_journals'         => envNonEmpty('SEND_REPORT_JOURNALS', true),
     'analytics_id'                 => env('ANALYTICS_ID', ''),
--- a/routes/web.php
+++ b/routes/web.php
@@ -21,6 +21,10 @@
 
 declare(strict_types=1);
 
+if ('' !== config('firefly.baseurl_override') {
+    URL::forceRootUrl(config('firefly.baseurl_override'));
+}
+
 Route::group(
     ['namespace' => 'FireflyIII\Http\Controllers\System',
      'as'        => 'installer.', 'prefix' => 'install'], function () {


JC5:
No, this doesn't feel right. I just know every idiot with a reverse proxy is going to use this to skip proper configuration. I know the issue with subfolders but this isn't the answer.

TimCinel:
For a service that doesn't make use of relative URLs 100% of the time, providing the root URL of the site *is proper configuration*. It's the standard solution to this problem, you see it in [Wordpress](https://stackoverflow.com/questions/28346286/why-is-my-wordpress-site-redirecting-to-incorrect-url), [NextCloud](https://docs.nextcloud.com/server/15/admin_manual/configuration_server/reverse_proxy_configuration.html), and more.

Correct me if I'm wrong, but as far as I'm aware there are two options for supporting a reverse proxy configured on a non-root path:

* The application always use relative paths, with no exceptions. This also means no HTTP redirects. [Syncthing works this way](https://github.com/syncthing/syncthing/issues/1106#issuecomment-66365527)), which is why there's no need to configure base URL for SyncThing reverse proxying.
* The application does not always use relative paths, so it must be aware of its external URL (site url, base url, whatever.). Wordpress works this way, which is why the site URL is one of the first options you configure during setup. Firefly III works this way, which is why this PR should be merged =)

There is another option. Instead of fixing it at the source, another option is to configure the proxy to rewrite HTTP headers and HTTP response content. That the least good solution, though. It's really just a janky workaround and it would be ideal if people [didn't have to resort to host Firefly III](https://github.com/firefly-iii/firefly-iii/issues/1869#issuecomment-450594411).

JC5:
Thanks. It still feels hacky. Especially coming from Wordpress. I'll do some research.

TimCinel:
> Especially coming from Wordpress

Hehe, I'm not going to try defending this one.

But yeah, I'm *relatively confident* that it's necessary whenever absolute URLs are used by an application (html, js, css, or even response headers).

JC5:
Yeah I know. But every moron (sorry, "less knowledgeable user") uses this feature or a similar one to leave their reverse proxy wide open or configure it terribly.

At least without this feature they'll have to think about before they act. Not being able to support /subdirectories is not a terrible price to pay.

TimCinel:
After trying to use this configuration more thoroughly I have found that the app doesn't always make use of routes for its redirects (particularly after creating/deleting entities).

That means non-root reverse proxy is still a not properly supported, even with this current PR. Given your hesitance towards this kind of config, I won't pursue any further code changes for this issue.

Guess I'll be setting up a new DNS record and configuring let's encrypt for it. Sigh.

Thanks again for considering.

Odyno:
My 2 cents... and maybe from your "less experienced user" point of view... You've done a great job, but by not introducing this feature in the real world you're only connecting to a third level name ( like http://finance.odyno.com ) because unfortunately it's not a feature and power for a second level ( like https://www.odyno-finance.com ) and you also preclude and prevent the possibility to add other tools on that server ( like nextcloud and similar)... too bad this tool will miss this opportunity.

BDW, the project is yours, I hope you will reconsider it one day

JC5:
You’re not making any sense to me. Please open a new ticket to explain what you mean.

pujan14:
@JC5 I'm also facing issue with subfolder. I believe this PR would fix the issue.

Currently I run multiple services under single subdomain like pujan14.example.com with each having their own subfolder.

Without support for subfolder I can not use this app and I would be forced to choose between not using this app and setting up another subdomain dedicated for this app. I personally would go with not using this app and I would strongly urge you to allow this PR or introduce fix for this issue.

Issue #2601, opened 2019-09-14
FinTs PSD2


Schikri:
I used to import via fints quite well with different banks.

Since this week my import does not work for any bank anymore. I suppose there is any other parameter needed beside of URL, Bankcode, Username and PIN.

Does anyone has an idea?

The errosr is: Request Failed: Unknown error. There is also nothing in the log files. I am using firefly version 4.7.17

JC5:
I'll need some logs to see what's wrong I'm afraid. What makes you say it's PSD2 related?

JC5:
Hi! This is a canned reply that tells you how to get more debug details.

1. Please collect some log files. Go to `storage/logs` and open the latest log file. It should be called something like "ff3-fpm-fcgi-*.log". In it, you should be able to find the error along with its "stacktrace": this will tell me where Firefly III broke down. If you run Docker, use `docker logs <container>`.
2. You may also [enable debug mode](https://github.com/firefly-iii/help/wiki/Enable-debug-logging-and-debug-mode) to make sure the log files catch everything and try again.
3. Make sure you clear your browser cookies and clear Firefly III's cache files by running `php artisan cache:clear` or by visiting `/flush` on your installation.

If you do not want to share log files or other details on Github, please [send them to me over (encrypted) email](https://firefly-iii.org/contact-information.html).

Schikri:
Here is my debug log:

[2019-09-15 17:46:11] local.DEBUG: Going to try to decrypt users currency preference.  
[2019-09-15 17:46:11] local.DEBUG: Could not decrypt "EUR". The payload is invalid.  
[2019-09-15 17:46:11] local.DEBUG: Now in postConfigure() {"job":"ijgjgYVGU4z4"} 
[2019-09-15 17:46:11] local.DEBUG: Going to create class "FireflyIII\Import\JobConfiguration\FinTSJobConfiguration"  
[2019-09-15 17:46:11] local.DEBUG: Updating configuration...  
[2019-09-15 17:46:12] local.DEBUG: Will NOT set previous URL to http://firefly-iii.idoleo.de/import/job/configuration/ijgjgYVGU4z4  
[2019-09-15 17:46:12] local.DEBUG: Now in JobConfigurationController::index()  
[2019-09-15 17:46:12] local.DEBUG: Now in JobConfigurationController::index() with job "ijgjgYVGU4z4" and status "has_prereq"  
[...]

PR #2622, opened 2019-09-18
Add support for new PSD2 regulations in FinTS import

luqasn:
Fixes issue #2601

Changes in this pull request:

- Switch to `nemiah/php-fints`, fork of `mschindler83/fints-hbci-php` with support for new PSD2 changes
- add config setting for fints registration number (needs to be filled out correctly)

Tested for these banks:
- [x] Sparkasse
- [x] DKB

Using this product registration: https://github.com/willuhn/hibiscus/blob/master/src/de/willuhn/jameica/hbci/HBCI.java#L218

@JC5


composer.json
```diff
@@ -83,7 +83,7 @@
     "league/flysystem-sftp": "1.*",
     "league/fractal": "0.*",
     "litipk/flysystem-fallback-adapter": "0.*",
-    "mschindler83/fints-hbci-php": "1.*",
+    "nemiah/php-fints": "dev-master",
```
luqasn: Unfortunately, there is still no release that supports this, but we might want to at least pin a commit?

config/import.php
```diff
@@ -168,5 +168,8 @@
         'plaid'   => [],
         'quovo'   => [],
         'yodlee'  => [],
+        'fints'   => [
+            'registration_number' => 'TODO'
```
luqasn: This obviously also needs to be changed before merging :)

JC5:
I'm picking this up in new export routines. 

https://www.patreon.com/posts/30012174

PR #2678, opened 2019-10-02
Add cron example in docker-compose.yml

bestlibre:
Changes in this pull request:

- add the example (commented) for the cron service from the documentation in docker-compose.yml

@JC5

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,6 +30,11 @@ services:
       - firefly_iii_net
     volumes: 
       - firefly_iii_db:/var/lib/postgresql/data
+  # uncomment the following lines after setting your Firefly IIU url and token to activate cron support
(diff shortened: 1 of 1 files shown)


JC5:
Add something to the documentation, not the YAML file please :)

bestlibre:
I should have open an issue first, sorry.

I will try to add a link in the [docker doc](https://firefly-iii.readthedocs.io/en/latest/installation/docker.html) to point to the docker cron.

But I think that having it in the yaml can be helpfull too. This way way you download the compose file you have (almost) everything needed.

JC5:
Thanks, please do.

Any tip, guide, solution or hint will lose its relevance over time. The software is patched, the bug is fixed, the problem is gone, etc. When they are all over the place it's hell to maintain.

PR #2707, opened 2019-10-09
Use the --show parameter to show the key

danger89:
Instead of modifying Laravel framework, only show the generated key (properly this is what you want?)


danger89:
Well I do think this is apparently exactly what you want.. However, of-course you can't run this script if you are using the Docker setup guide. Uhm maybe this PR should be closed.

Instead, update the docs?

JC5:
As requested.

PR #2709, opened 2019-10-10
Minor bugfix on Rule editing

dguhl:
Changes in this pull request:

- updated .gitignore to ignore JetBrains IntelliJ/PhpStorm config directory
- added a secure key generation to entrypoint.sh, right after (Docker Container) installation
- After editing a rule (Money Management / Rules), the application should only return to the view when the user checked `return_to_edit`, but when it's not checked, the logical path would be to return to the Rule index (from which a user usually came when editing a rule)

The function `previousUri` only checks for previous called URIs in the section (blacklisting certain destructive actions), which is a good idea when the previous views always have been stored in a history on the session. For the `index` view, I was not able to make out the call which stores the URI into the session, so `previousUri` was not able to return to the previous point.

_Sidenote:_ I accidentially committed my changes to my fork's `master` branch, but I opened the PR from that to your `develop` branch.

@JC5


JC5:
1. It's a best practice to put those in your global `.gitignore` file.
2. Key generation is pointless since it will be overruled by whatever you submit using environment variables. And since the `.env` file disappears when your rebuild the container it might not be a good place to store the key.
3. I think a better and more flexible fix would be to make sure that the index *does* store the previous URL. This makes the edit page redirect more flexible, since it can also be reached from other pages.

JC5:
The key is also used to encrypt data, so resetting it every time seriously messes up your administration. I’m open to suggestions to improve key generation but creating a new one isn’t the answer. 

As for the rule suggestion, and the history thing, feel free to open an issue so we can discuss it.

dguhl:
> The key is also used to encrypt data, so resetting it every time seriously messes up your administration. I’m open to suggestions to improve key generation but creating a new one isn’t the answer.

I see. But I searched the code (outside vendor) for where encryption __involving this key__ would happen, and I didn't find anything but configuration values to activate encryption. But it seems like there's no place in the code where this configuration is actually checked.

> As for the rule suggestion, and the history thing, feel free to open an issue so we can discuss it.
Great. We maybe should take out this specific problem from this PR altogether and do it the right way from the first place, separately.

dguhl:
The `key:generate` command rejects (or prompts to confirm) to overwrite the key when the environment is `production`. Unfortunately, that's not the case with Firefly III by default. Anyways, ever since commit `bc735e3a59afc463c0bd9a055fda29895dd792b8`, there is no `.env` file in the container at all when it is generated.

PR #2710, opened 2019-10-10
Improve ordering

danger89:
Worst docker yml ordering I have ever seen. So I fixed the layout and make it readable :+1:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,4 @@
---- 
-networks: 
-  firefly_iii_net: 
-    driver: bridge
+version: "3.2"
 services: 
   firefly_iii_app: 
     image: jc5x/firefly-iii:latest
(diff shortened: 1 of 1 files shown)


JC5:
I don't think users need to see or manage this file. I appreciate the addition of `type: volume` for the first volume but the rest I really don't think I need.

danger89:
People will see and manage this file. This is the Docker yml configuration file. Administrators who are setting up Firefly using Docker will use this file!

Content is the same, but it is much more readable, using the template which is often used:
https://docs.docker.com/compose/compose-file/

Eg. `version` is at the top. `Networks` and `volumes are below the `services`, etc etc. This will help people.

JC5:
They’ll be fine.

danger89:
:disappointed:

PR #2722, opened 2019-10-11
Add IBAN & Clean up status v2

danger89:
I really would like to see the IBAN number next to the account name on the Account->Show page.

Currently if I click on the account name, I have no clue what is the matching IBAN number.

Moverover, I made the status polling spam the console less.

I hope you like it!

Regards,
Melroy


JC5:
Looking good. I got to make a few points though.

* Please don't submit translations, [Crowdin](https://crowdin.com/project/firefly-iii) will pick these up.
* Users can also have no IBAN, an account number or both. I see no provisions for this.
* Could you split the console thing into another PR?

JC5:
And personally, cramming the IBAN between brackets to make it visible feels a bit sloppy. Let me think of something as well.

danger89:
> Please don't submit translations, Crowdin will pick these up.

Was not my intention, but **needs** to be changed together with this change, otherwise it won't work.
> Users can also have no IBAN, an account number or both. I see no provisions for this.

I agree, currently there is no check about the name as well 😕 

> Could you split the console thing into another PR?

Sure, was thinking the same thing.

JC5:
About crowdin. I have to push then pull and translators need to pick it up. That’s the price of progress.

danger89:
> I have to push then pull and translators need to pick it up. That’s the price of progress.

If you look closely, this doesn't need translations. The only reason I needed to adapt those files, are for showing the IBAN.

JC5:
Yes I know that but that’s not how the system works.

PR #2807, opened 2019-11-07
Fix account endpoint in create-edit.js

timendum:
Changes in this pull request:

-    change the endpoint for creating/editing rules to the correct one

This issue affects also current stable release.

@JC5

This should be a legitimate pull.

Also a warning: all "convert" select list suffers from the same problem, but i can't find a way to create an autocomplete for accounts filtered by type.

`createAutoComplete(inputResult, 'json/accounts?type=revenue');` does not works.

A temporary workaround should be `createAutoComplete(inputResult, 'json/accounts');`  but does not filter the account type.


JC5:
Thanks. I'll wait with merging to remind myself to pick this up.

JC5:
Merged, and fixed the missing asset accounts.

PR #2932, opened 2019-12-30
Change a newly-deprecated array access

pascalj:
PHP allowed array accesses with curly braces but [deprecated the feature](https://wiki.php.net/rfc/deprecate_curly_braces_array_access) with 7.4. With this change the CSV import will work on 7.4 without any errors or warnings. The change does not affect other versions.

Side node: everything else I tested seems to work well with 7.4.1.


JC5:
Nice fix, merged.

PR #3094, opened 2020-02-07
Locale on Windows

wzzirro:
Changes in this pull request:

- Fixed currency locale on Windows

On Windows, setlocale(LC_ALL, '') sets the locale names from the system's regional/language settings


JC5:
Thanks!

The goal of the code where you made your change is to change to the locale in `trans('config.locale')`. 

Your code however, would result in different behavior on Windows, always going back to whatever Windows is configured for.

So it seems to be doing the wrong thing.

PR #3097, opened 2020-02-08
API: Unified attachment_* and model_* for Attachments

GeoffreyFrogeye:
Currently when creating/updating an Attachment, we need to use the fields `model` and `model_id` to specify the object the attachment refers to. However to read the same values back from the API the names are `attachable_type` and `attachble_id`.

This PR fixes this inconsistency. I choose to roll with `attachable` rather than `model` because it makes more sense and (probably) requires the least changes, even though it creates an inconsistency with the DB table.


JC5:
Nice work as always :+1:

Issue #3091, opened 2020-02-07
Redis ERR invalid DB index


Agraphie:
I am running Firefly III version 5.0.4 in docker.

**Description**
I'm trying to use redis and get the following error
```
In AbstractConnection.php line 155:
                                                            
  `SELECT` failed: ERR invalid DB index [tcp://redis:6379]  
```

**Extra info**
This is my config in .env
```
# If you're looking for performance improvements, you could install memcached.
CACHE_DRIVER=redis
SESSION_DRIVER=redis

# If you set either of these to 'redis', you might want to update these settings too
# If you use Docker or similar, you can set REDIS_HOST_FILE, REDIS_PASSWORD_FILE or
# REDIS_PORT_FILE to set the value from a file instead of from an environment variable
REDIS_HOST=redis
REDIS_PASSWORD=null
REDIS_PORT=6379
REDIS_DB="0" # always use quotes
REDIS_CACHE_DB="1"
```

Redis is reachable from within the container on port 6379 though.

Any help is appreciated :)

JC5:
You could try with `0` instead of `"0"`. How many databases is redis configured with?

Agraphie:
Hi,

thanks for your (very fast :D) reply. I tried `0` (without quotes) as well, but got the same error. Redis has one database.

```
127.0.0.1:6379> info keyspace
# Keyspace
db0:keys=66,expires=15,avg_ttl=768690
```

So theoretically `"0"` seems fine. Unfortunately, I'm a bit of a redis noob so I wouldn't really know what to look for.

JC5:
You'll need at least two, one for the cache as well.

Agraphie:
That's true, but I also set the "REDIS_CACHE_DB" to `0` and I got the same error. Technically it should be fine to run in one db, right?

JC5:
I think it should be yes, but to be honest I'm not sure.

I ran into similar issues when I started to use redis but honestly I can't recall what I did to fix it. Let me look it up later.

Agraphie:
I changed the small thing seen in the PR above and it started working for me. Maybe you had the same issue back then?

Agraphie:
I would close this issue @JC5 as it seems to be resolved. Otherwise just reopen if you think something is missing :)

JC5:
I think so yes. Thanks for the feedback! The issue can remain closed :+1:

PR #3100, opened 2020-02-08
Move redis comment to different line

Agraphie:
Fixes issue #3091 

Changes in this pull request:
Move comment about quotes in redis db index to own line. The comment in the variable caused redis to use `"0" #always use quotes` as redis db index instead of just `"0"`.


JC5:
:+1:

PR #3104, opened 2020-02-09
Fixed Transaction page attachment broken due to #3097

GeoffreyFrogeye:
I though https://github.com/firefly-iii/firefly-iii/pull/3097 would break nothing but I forgot the transaction page relied on the API. So I fixed it.

I used `yarn prod` to generate the Javascript in `public`, is that the right thing to do? I did not find any documentation about it.


JC5:
That's correct. I might regenerate it anyway because of another attachment related issue. Thanks for the fix!

GeoffreyFrogeye:
Ok, thanks!

If I happend to fix another thing like that, do you want me to commit the regenerated JS or do you prefer me to generate it yourself when needed in order to keep the size of the git repo at a minimum?

JC5:
Not really sure. But now that you mention it, forego the commit. I always run it before a version release anyway.

SuperSandro2000:
@GeoffreyFrogeye I would prefer if the JS was always commited to be be able to run that commit without strange issues.

PR #3131, opened 2020-02-18
Fix search modifiers claiming the whole query

cpmsmith:
The regex being used to detect search modifiers was only restricted to characters without the Unicode property Other, which is most characters, including spaces. That meant that almost anything following a modifier got treated as part of it, breaking, among other things, combining modifiers. This version will include anything but a space, unless the value is quoted, in which case it'll accept anything.

### Old regex:
Screen Shot 2020-02-17 at 20 25 56

### New regex:
Screen Shot 2020-02-17 at 20 27 41


JC5:
👍

PR #3137, opened 2020-02-21
add 2 <th> to fix header row border, fix alignment on action btn

DanYoSon:
Changes in this pull request:
- Bottom border on transactions list header did not continue the whole way, a couple of column headers were missing
- action buttons were aligned differently for single and split transactions.

**Before:**
Firefly-Before-Annotated
**After:**
Firefly-After

@JC5 didn't discuss first hope that's ok.


JC5:
That's quite alright, thanks! :+1:

Issue #2642, opened 2019-09-22
[ING] Remove name from description


rubenverhoef:
**Description**
I would like to have the `Naam: xxxx` to be removed from the description field for a ING  `csv` import. This because this data is already at column `2`.
I see the `IBAN:` is already removed which is nice because that is also already in a column.
With the `Naam:` being removed it should map my duplicates from different accounts, see #1863

So my second question is, does the import check for duplicates before or after the special ING "fix"?

JC5:
That may happen if somebody can write it, in the new CSV import routine for Firefly III. You can read about this on [Patreon](https://www.patreon.com/posts/30012174). To answer your question: Firefly III will check for duplicates after the ING fix.

rubenverhoef:
Okay, when the new importer is there i have a look to add this. 

I looked at this code and it should be removed:
https://github.com/firefly-iii/firefly-iii/blob/d550a626e90e6307603d12fa36241fbc98bf0be3/app/Import/Specifics/IngDescription.php#L74-L86
I looked further and the problem is that some of my entries are not having the `Omschrijving`:
https://github.com/firefly-iii/firefly-iii/blob/d550a626e90e6307603d12fa36241fbc98bf0be3/app/Import/Specifics/IngDescription.php#L118

So when the entry has no `Omschrijving` the `Naam:` entry will not be removed.

I will look if i can fix this issue and create a pull request, or do you have a quick fix for this?

JC5:
If you do, please do so on this file:

https://github.com/JC5/firefly-iii-csv-importer/blob/master/app/Services/CSV/Specifics/IngDescription.php

And send me some example lines over mail so I can check if it works as expected.

rubenverhoef:
I'm working on it and it goes fine, few questions about the current method:

what does this function do?:
https://github.com/firefly-iii/firefly-iii/blob/d550a626e90e6307603d12fa36241fbc98bf0be3/app/Import/Specifics/IngDescription.php#L124-L130
Do you have an example and i think it is not working for every situation because my savings acount is named `Bonusrenterekening`

What i now have:
* Remove `Naam` entry from description (doesn't matter what is behind or before it)
* Remove `Omschrijving:` but not the value of it. So only the text `Omschrijving`.
* copy `Valutadatum` to own column so we can use it as a FireFly iii parameter and it makes the "omschrijving" much cleaner.

What i will make:
[...]

PR #3146, opened 2020-02-23
Better CSV import for ING (NL)

rubenverhoef:
Fixes issue #2642

Changes in this pull request:

- Added `DV` transaction type (Divers)
- Correctly remove `Naam` and it's value from `Mededelingen` collumn
- Remove the word `Omschrijving` from `Mededelingen` collumn
- Fix copy of bank account number and name when transaction is an transfer from/to saving account.
- Move `Valutadatum` from `Mededelingen` collumn to an own collumn. Right now the collumn itself has no name.

Todo:
- Add an name for the newly created collumn. I have done this hardcoded by adding this line:
``` php
array_push($headers, "Valutadatum"); // New row for "Valutadatum" (only for IngDescription.php specific import)
```
To here (between `$headers` and `// get example rows:` (line 253))
https://github.com/firefly-iii/firefly-iii/blob/af503597e62108cd7fedaa8ef89c80d161cbf304/app/Support/Import/JobConfiguration/File/ConfigureRolesHandler.php#L250-L255

@JC5 Do you have any idea how i can add an name for the newly created column in a more general way and only for the ING specific task?

### DIFF
--- a/app/Import/Specifics/IngDescription.php
+++ b/app/Import/Specifics/IngDescription.php
@@ -80,8 +80,7 @@ class IngDescription implements SpecificInterface
                     $this->removeIBANIngDescription();   // Remove "IBAN:", because it is already at "Tegenrekening"
                     $this->removeNameIngDescription();   // Remove "Naam:", because it is already at "Naam/ Omschrijving"
                     $this->removeIngDescription();       // Remove "Omschrijving", but not the value from description
-                    // if "tegenrekening" empty, copy the description. Primitive, but it works.
-                    $this->copyDescriptionToOpposite();
+                    $this->MoveSavingsAccount();         // Move savings account number and name
                     break;
                 case 'BA':                              // Betaalautomaat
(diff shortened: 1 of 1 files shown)


JC5:
Thanks! I'll dive into the code asap. I'm probably going to haul it over into this file as well:

https://github.com/firefly-iii/csv-importer/blob/master/app/Services/CSV/Specifics/IngDescription.php

rubenverhoef:
Yeah, it has to be also included in the [csv-importer](https://github.com/firefly-iii/csv-importer).

The only thing is the name of the newly created column, we have to find an more generic way to do it.

JC5:
See here:

https://github.com/firefly-iii/csv-importer/blob/develop/app/Services/CSV/Specifics/IngDescription.php#L168

https://github.com/firefly-iii/csv-importer/blob/master/app/Services/CSV/Specifics/SpecificInterface.php#L39

PR #3197, opened 2020-03-18
Fix small issue from #3146

rubenverhoef:
See #3146 there was a small issue when there was no `valutadatum` in the `omschrijving`

@JC5


JC5:
Nice, merged.

PR #3206, opened 2020-03-25
fix parameters on budget

juhniorsantos:
adjustment in syntax, was breaking the application.

@JC5


JC5:
This is an allowed construct in PHP7.3. 


https://php.watch/versions/7.3#func-call-trailing-comma

Issue #3277, opened 2020-04-19
Is updating the layout an option?


lguima:
I am running Firefly III version 5.1.0.

**Description**
I would like to suggest a rearrange for the menu, but then I figure out the version of AdminLTE is being used by Firefly III support only Font Awesome 4.7.0 that have a limited set of icons.

Searching here in issues, I saw @JC5 mentioning some time ago the AdminLTE upgrade is a pain.

But I also saw in the [Projects](https://github.com/firefly-iii/firefly-iii/projects/5) tab there's some issues under a section called "As part of new layout".

**Is this outdated or is a work in progress?**

If this is a work in progress, where can I find more details on how can I contribute?

Or if this is outdated, is an option to update the layout to the latest version of AdminLTE?

Or even to a new one, like the [SB Admin 2](https://startbootstrap.com/themes/sb-admin-2/) from [Start Bootstrap](https://startbootstrap.com/), the same source as the Firefly III website?

JC5:
It's not a priority. Any effort for a new layout will probably be a SPA. What did you have in mind?

lguima:
I was looking for a layout that supports more icons to implement the menu rearrange suggestion.

Searching for AdminLTE as a SPA, I found some projects with Vue.js and React, but outdated.

And the theme [SB Admin 2](https://startbootstrap.com/themes/sb-admin-2/) mention in their feature it was built heading on that:
- _Minimal use of jQuery and JavaScript to make it easier to rewrite using JS frameworks_

**Do you think the effort to upgrade AdminLTE or update to SB Admin 2 is the same as to implement as a SPA?**

JC5:
What was your suggestion?

lguima:
It is to group some features and reorder them in order to encourage the use (Financial Control and Automation) and make more accessible by the frequency of use (Transactions).

And placing the meta configuration and settings related to the user under the user options.

**Menu sample:**
![firefly-menu](https://user-images.githubusercontent.com/1562097/79701553-67f2bd00-8274-11ea-9a89-50b0ab5f224c.PNG)

**Menu transcript:**

- Dashboard
---
_Financial Control_
- Budget
- Bills
- Piggy Bank
---
_Accounting_
- Transactions
  - Expenses
  - Revenue/Incomes
  - Transfers
- Automation
  - Recurring transactions
  - Rules
---
- Accounts
  - Asset accounts
  - Expense accounts
  - Revenue accounts
  - Liabilities
---
- Classification
[...]

PR #3305, opened 2020-04-28
Menu reordering

lguima:
It's related to the issue #3277

Changes in this pull request:

- Reordered and grouped some items in the sidebar menu according to their frequency of use and to encourage the use of them

@JC5

### DIFF
--- a/resources/views/v1/layout/default.twig
+++ b/resources/views/v1/layout/default.twig
@@ -86,27 +86,9 @@
 
                     {% if not SANDSTORM %}
                         <li class="dropdown user user-menu">
-                            <a href="#" class="dropdown-toggle" data-toggle="dropdown">
-                                <span class="visible-xs">@</span>
-                                <span class="hidden-xs">{{ Auth.user.email }}</span>
-                            </a>
-
-                            <ul class="dropdown-menu" style="width: unset;">
-                                <li class="user-footer">
-                                    <div class="col-xs-6 text-center">
-                                        <a href="{{ route('profile.index') }}" class="btn btn-default btn-flat">
-                                            <span>{{ 'profile'|_ }}</span>
-                                        </a>
-                                    </div>
-                                    {% if not SANDSTORM %}
-                                        <div class="col-xs-6 text-center">
-                                            <a href="{{ route('logout') }}" class="btn btn-default btn-flat">
-                                                <span>{{ 'logout'|_ }}</span>
-                                            </a>
-                                        </div>
-                                    {% endif %}
-                                </li>
-                            </ul>
+                        <span style="cursor:default;color:#fff;padding: 15px;display: block;line-height: 20px;">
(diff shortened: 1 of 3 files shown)


lguima:
@JC5 
To create a string translatable, should I place it in all language files or only in the en-US?

JC5:
Putting the strings in en_US is enough.

resources/views/v1/layout/default.twig
```diff
@@ -86,9 +86,27 @@
...
-                        </span>
+                            <a href="#" class="dropdown-toggle" data-toggle="dropdown">
+                                <span class="visible-xs">@</span>
+                                <span class="hidden-xs">{{ Auth.user.email }}</span>
+                            </a>
+
+                            <ul class="dropdown-menu" style="width: unset;">
+                                <li class="user-footer">
+                                    <div class="col-xs-6 text-center">
+                                        <a href="{{ route('profile.index') }}" class="btn btn-default btn-flat">
```
JC5, line 97: I'm not a fan of buttons here. They render badly and it's unexpected UI. Can this be a dropdown list?

JC5:
Dropping the icons for the submenu's has the weird effect that the sub-items are more to the left than the main items. I suggest you leave the icon for the sub-items.

Budget icon is OK. Use the empty calendar version for bills. Don't use the hamburger menu icon for options. Maybe [sliders](https://fontawesome.com/v4.7.0/icon/sliders)?

resources/views/v1/layout/default.twig
lguima: Actually, this is from [the official main header template](https://adminlte.io/docs/2.4/main-header), what is painful to work without the original layout.

Also it broken when I was taking screenshots and replace my email address to a smaller one.

Then, I'm reverting, I'll left in the sidebar menu.

lguima:
> Budget icon is OK. Use the empty calendar version for bills. Don't use the hamburger menu icon for options. Maybe [sliders](https://fontawesome.com/v4.7.0/icon/sliders)?

Agreed, I updated the icons.

> Dropping the icons for the submenu's has the weird effect that the sub-items are more to the left than the main items. I suggest you leave the icon for the sub-items.

Indeed, but the Font Awesome 4.7 is very limited, I also used some of the previous sub-menu icons for some menu items, then it will cause duplications.

But looking for the official live preview they use a [circle-o](https://fontawesome.com/v4.7.0/icon/circle-o) for each sub-menu item, but it doesn't seems good to me.

Then I tried with [angle-right](https://fontawesome.com/v4.7.0/icon/angle-right) and it seems better.

sidebar-menu

JC5:
Yeah that could work too.

lguima:
Great, thanks! I marked the PR as **Ready for Review** now.

JC5:
:+1:

Issue #3322, opened 2020-05-01
Pages and blocks icons outdated


lguima:
**Bug description**
I am running Firefly III version `develop`, and my problem is:

I worked on the menu reordering but totally forgot to update the icons around the site, on the top of the pages and on the blocks related.

I'm opening this issue to track until I or someone can fix.

**Steps to reproduce**
- Access a page that had the icon changed in sidebar menu and compare the changed icon with the one on the top of the page.
- Access a page that have a block related to the pages that had the icon changed and compare the changed icon with the one on the block.

JC5:
No worries, I already fixed some of them, the rest will follow in due course. Since this is still the develop branch, no need to panic.

lguima:
Got it, just in case you were planning a release.

JC5:
Yes, but nothing major. I noticed that the Transactions overview page says "Accounts" instead of "Transactions" and nobody noticed (including me) for over 8 months. We're good :wink:

PR #3323, opened 2020-05-01
Update the pages and blocks icons

lguima:
Fixes issue #3322

Changes in this pull request:

- Update Budgets pages and block
- Update Piggy Banks pages
- Update Categories pages and block
- Update Tags pages
- Update Reports pages

@JC5


lguima:
I thought it would be heavier to fix, but there it is.

- I updated the Categories icon because it was conflicting with the new Reports icon.
- And also updated the Tags icon because it was conflicting with the Classification icon.

The Categories icon [bookmark](https://fontawesome.com/v4.7.0/icon/bookmark) was the best I could find, but let me know if you have any other suggestion.

JC5:
Looks good, nice work :+1:

lguima:
Alright, then it's ready for review.

Issue #3227, opened 2020-04-01
Add SSL config for MySQL and LDAP


bpatath:
**Description**
I need to use SSL/TLS to connect to the Mysql database. This is supported by Laravel using the `options` key in the database definition and standard PDO options:

```php
'mysql' => [
    ...,
    'sslmode' => 'prefer',
    'options'   => array(
        PDO::MYSQL_ATTR_SSL_CAPATH => '/.../certs/',
        PDO::MYSQL_ATTR_SSL_CA     => '/.../ca-cert.pem',
        PDO::MYSQL_ATTR_SSL_CERT   => '/.../cert.pem',
        PDO::MYSQL_ATTR_SSL_KEY    => '/.../key.pem',
        PDO::MYSQL_ATTR_SSL_CIPHER => 'TLSv1.2 TLSv1.3',
        PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT => true,
    ),
];
```

The full list is documented [here](https://www.php.net/manual/fr/ref.pdo-mysql.php#pdo-mysql.constants), but these are the most commonly available on SSL/TLS compatible softwares.

Likewise, Adldap supports additional SSL/TLS options through the 'custom_options' key in the settings configuration:

```php
'settings' => [
    ...,
    'use_tls' = true,
    'custom_options' => array(
        LDAP_OPT_X_TLS_CACERTDIR    => '/.../certs/',
        LDAP_OPT_X_TLS_CACERTFILE   => '/.../ca-cert.pem',
        LDAP_OPT_X_TLS_CERTFILE     => '/.../cert.pem',
        LDAP_OPT_X_TLS_KEYFILE      => '/.../key.pem',
        LDAP_OPT_X_TLS_CIPHER_SUITE => 'TLSv1.2 TLSv1.3',
        LDAP_OPT_X_TLS_REQUIRE_CERT => LDAP_OPT_X_TLS_HARD,
    ),
];
```

The full list of options is available [here](https://www.php.net/ldap_set_option)

**Solution**
Add environment variables for each of these options, and generate corresponding arrays in `config/database.php` and `config/ldap.php`. I'm willing to do a PR.

JC5:
Thanks, I'll pick it up.

PR #3393, opened 2020-05-24
Add SSL conf to MySQL and LDAP

bpatath:
Fixes issue #3227 

Changes in this pull request:

- Add environment variables to configure MySQL SSL options
- Add environment variables to configure LDAP SSL options

It is a bit ugly... but after some testing, using `null` or `""` as a default value breaks the SSL config. Only actually used options should be passed for it to work.

Also, I copied the "set from file by appending _FILE" comment, but do I need to do something to handle this case ?

@JC5


JC5:
Alright, this is OK. What do you mean exactly with this?

> It is a bit ugly... but after some testing, using null or "" as a default value breaks the SSL config. Only actually used options should be passed for it to work.

- I **must** set all six options to *something* in order for the config to work?
- Empty values are not allowed?

In that case, better if-statements to validate the content are necessary. Correct?

bpatath:
You must not set an option to anything other than a valid value. Mysql and LDAP will use default values for an option only if its corresponding key does not exist in the `options` array. Setting the option to `null` or `""` are not considered valid values, nor placeholder for "use the default Mysql/ldap value".
Thus, the following would work only if the six options are provided with valid values:
```
'mysql' => [
    ...,
    'sslmode' => 'prefer',
    'options'   => array(
        PDO::MYSQL_ATTR_SSL_CAPATH => envNotEmpty('MYSQL_SSL_ROOT_CERT_DIR', null),
        PDO::MYSQL_ATTR_SSL_CA     => envNotEmpty('MYSQL_SSL_ROOT_CERT', null),
        PDO::MYSQL_ATTR_SSL_CERT   => envNotEmpty('MYSQL_SSL_CERT', null),
        PDO::MYSQL_ATTR_SSL_KEY    => envNotEmpty('MYSQL_SSL_KEY', null),
        PDO::MYSQL_ATTR_SSL_CIPHER => envNotEmpty('MYSQL_SSL_CIPHER', null),
        PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT => envNotEmpty('MYSQL_SSL_VERIFY', null),
    ),
];
```

That's why I had to use consecutive if statements to incrementally build the `options` array, with only options actually provided by the user.

From a user perspective, it's normal. Environment variable unset or set to empty string means the corresponding SSL option will not be set. Any non-empty value means the option will be set to this value.

JC5:
Alright, so it's all or nothing. Could you do something like this to make the code easier? It would also allow the user to enable / disable the feature.

(bit of pseudocode im lazy)

```
configarray = [];
if(MYSQL_SSL_MODE == require) 
  // set up ssl array
configarray = [bla bla]
}
```

It would be safer as well because MySQL would not fall back to non-SSL if it fails.

bpatath:
Ok sorry, I'm not clear. :(

The code in the PR is the good one. I was just trying to explain the rationale behind using consecutive if statements to fill the options arrays, vs using a static construct as in the rest of the file and my previous comment.

- Users may want to use any combination of the SSL options. They may use none, all of them, or a subset of the options.
- They may also use them with any SSL mode, not necessarily required.
- Only the SSL mode can prevent Mysql/ldap from falling back to a non-secure connection.
- Both Mysql and LDAP will blindly use the value of the option *iff its key exists in the array*, independently of the actual value. Setting an option to `null` is not equivalent to not setting it. To not use an option, its key should not be present in the array.
- Thus, the only option is to build the array so it contains only keys corresponding to options the user wants to use

JC5:
Thanks for the explanation. What I meant to say was: why not build the array only if the user sets the sslmode to `require`?

bpatath:
After digging a bit, although it exists in the MySQL CLI client, there is not such thing as an SSL mode in the PHP MySQL driver... I blindly used a snippet I found on a Laravel forum. It is crazy how undocumented all of this is, I had to check the C code of the driver.

MySQL CLI modes are: `disabled`, `preferred`, `required`, `verify_ca`, `verify_identity`.
The driver only supports equivalents of `disabled`, `required` and `verify_identity`.
If defaults to not using SSL. If any of the SSL options (except `MYSQL_ATTR_SSL_VERIFY_SERVER_CERT`) is set, it will use SSL with or without verification depending on the `MYSQL_ATTR_SSL_VERIFY_SERVER_CERT` value.

We can either:

- remove the `MYSQL_SSL_MODE` variable.
- replace it to a boolean `MYSQL_USE_SSL` which will not be passed to PDO but used to build/not build the options array, as suggested in your last comment. Going this way is maybe a bit clearer, and more coherent with pgsql and LDAP behaviors.

For LDAP, there are explicit `USE_SSL` and `USE_TLS` options. If both are false, SSL options are ignored so we can leave the code as is.

JC5:
I like the idea of something like `MYSQL_USE_SSL`. If people set it and expect SSL to magically work they're moron, and I don't cater to those. A little disclaimer in the `.env.example` file should be enough.

I would prefer `MYSQL_ATTR_SSL_VERIFY_SERVER_CERT` to be always set to true if people use SSL though, otherwise it's pointless.

bpatath:
Done. They might assume that without using any option, some defaults would be used. Anyway, I added a disclaimer, and an example value of a CA path for basic usage.

`MYSQL_ATTR_SSL_VERIFY_SERVER_CERT` defaults to `true`, but it is important to be able to disable server certificate verification when using a self-signed certificate.

JC5:
Merged!

sephrat:
This PR was mistakenly attributed to myself in the release note. @JC5 Could you please edit it and credit @bpatath for his work?

JC5:
Whoops! Will do, sorry @bpatath!

Issue #3387, opened 2020-05-23
Email notifications cannot be translated


sephrat:
I already reported this a few months ago in #3159 (item 6) and you said you wouldn't fix it.
Just wondering now the reason behind it? I thought it had to do with Twig templates not being easily translatable but I bumped into some twig files that are translatable.
Is it because it's too much work to do for little value to you (which I totally understand), or is there a technical issue behind it that I have overlooked?
In the former case, I'd like to give it a try if you don't mind. I'd start with the "Firefly III has created a new transaction" one. From what I see, for every email form I just need to change the related app/Mail/*.php file and the 2 associated twig files in firefly-iii/resources/views/v1/emails/, right? 
Are the -text.twig still used? I never saw an option to send plain text instead of HTML.

JC5:
To answer your last question first: Firefly III includes both in all emails, so proper clients fall back to the text-only view.

Laravel changed somewhat in the meantime and they can be translated. Keep in mind though that emails are sent separately from the user's session. In other words, you can't be sure you can find the user and their preferred language easily, from the perspective of an email message. That's the reason their not translated.

And I don't want to spagetti the user management code to the email code. Unless this improved in Laravel in the meantime, but honestly I havent spent much time on it yet.

sephrat:
How about we use default language for a start? Are there many Firefly III instances with users whose language differs from the default one? I can only think of people monetizing their Firefly III instance - is that a thing? And anyway if it were the case, those users would still have the login screen in the default language so they would already be used to it.

I'll play around with it and see what I come up with.

JC5:
Yes, Firefly III features a lot of languages but most users don't change the environment variables to overrule the language for all users.

PR #3395, opened 2020-05-25
Make email notifications translatable (#3387)

sephrat:
I made a POC with the admin test email form. It works fine. It uses the user's language out of the box for that form, which is nice. 
I ran a test with the "OAuth client created" notification: that one translates according to DEFAULT_LANGUAGE setting. I guess it will all depend on whether the email sending is done asynchronously of the user's session or not. 
To me this is acceptable behaviour. Drafting this pull request to get a second opinion.

TODO:
- [X] Header and footer
- [X]  Admin test
- [x] Access token created
- [x] Confirm email change
- [x] Error
- [x] Oauth client 
- [x] Password
- [x] Registered
- [x] Report new journals
- [x] Undo email change

Left as-is:
- [x] Confirm account (obsolete?)


JC5:
Nice work! Some quick feedback from my phone:

Move the translations to a separate file. Use {{ ipAddress }} and nothing else (should be a consistent variable)

JC5:
The mail code in /app/Mail uses the $ipAddress variable. You should be fine only using that one.

sephrat:
Regarding ipAddress, I just copy/pasted the initial code. I've cleaned it up like you said.
I'm almost done, should be finished by EOW. 
Any idea how I could test the text version? I've only been able to test html format with Mailtrap.

JC5:
Mailtrap should also display the text version on the other tabs.

sephrat:
It's all done now. I managed to test them all.
I left "Confirm account" as-is because it seems not to be used anymore. Maybe we should remove the files if that's the case?

JC5:
Indeed, that's an old one that's no longer used. I don't think I have write access to your branch, feel free to remove the files.

This looks amazing, and it was way less of a hassle than I feared. Good work! :+1:

sephrat:
Done!

JC5:
Nice work! 

I'm not sure what's etiquette in cases like this but I'd prefer this: `{{trans('email.email_change_old', {oldEmail: oldEmail})|raw }}` over this: `{{trans('email.email_change_old')|raw }} {{ oldEmail }}`. 

And I'd prefer it if HTML was part of the translation string, but I can see why you did that.

Should I keep asking you to change these things, or should I just merge it and do some changes myself? I really like what you did here, and I don't to give the impression it's wrong. What do you think?

sephrat:
> And I'd prefer it if HTML was part of the translation string, but I can see why you did that
I'm not sure I understand. Do you refer to the few times I passed html tags as variables instead of putting them in the strings? I did that to keep only one string so we don't have double maintenance for html and text. Unfortunately that makes the template kind of ugly. Would you rather split the strings in that case?
> 
> Should I keep asking you to change these things, or should I just merge it and do some changes myself? I really like what you did here, and I don't to give the impression it's wrong. What do you think?
No problem at all, that's basic peer review. I don't mind!

JC5:
> I'm not sure I understand. Do you refer to the few times I passed html tags as variables instead of putting them in the strings? I did that to keep only one string so we don't have double maintenance for html and text. Unfortunately that makes the template kind of ugly. Would you rather split the strings in that case?

I was thinking, maybe there's a strip_tags thing in Twig? Then you'd have the best of both worlds.

sephrat:
You're absolutely right, there's `|striptags` that does a fantastic job at it. I managed to merge some strings with that as well. Thanks for the tip.
This should be it now, please review.

PR #3404, opened 2020-05-27
Make new profile settings translatable

sephrat:
Just found that profile Vues weren't translatable. I took the opportunity to regroup i18n declarations so we can juste reuse it for any Vue.


sephrat:
BTW, no idea what's wrong with my yarn setup, the generated .js files look nothing like the original ones. Shouldn't it minify by default?

JC5:
It should yes! Thanks also for cleaning up the i18n code. You could also choose not to include the built files, I'll do them again on my side. I have to do so anyway to pick up the translations (I have a little script that does that).

JC5:
Made the following changes (see commits).

- Moved all strings from the 'profile' namespace to the 'firefly' namespace. 
- Added all new strings to en_US/firefly.php for translation, under the 'profile_' key.
- Pushed everything to Crowdin and downloaded the results (so the keys are up to date, if untranslated)
- Regenerated JS.

PR #3405, opened 2020-05-27
Make error pages translatable

sephrat:
I could not find a way to test 500 and 503 error pages.


JC5:
Browse to /error to test those. I’ll review later :)

sephrat:
> Browse to /error to test those.

/error does show FireflyException.twig, but how do I test 500 and 503 ? I tried /error/500 but this gives a 404.

JC5:
Mmmm. I'm just not sure of uncaught errors (which would cause the 500).

The 503 can be tested, I just realised, by turning on maintenance mode. `php artisan maintenance on` I think it is.

sephrat:
Right, I managed to test 503 with `php artisan down`. Thanks.
Will try to break things to get that 500 error page. Fun fact: if you throw an Exception in FireflyException.twig you get an infinite loop. :-)

JC5:
Nice! I've written down the infinite loop. Just to see if I can catch it somehow.

JC5:
Merged :+1:

PR #3413, opened 2020-05-29
String pluralization

sephrat:
I made the strings pluralization-friendly.
https://laravel.com/docs/6.x/localization#pluralization
Not a big thing for English language but grammar can be a mess in other languages (starting with mine). 

I went through all the strings thoroughly looking for "(s)" so I shouldn't have missed any. The remaining ones aren't relevant as they are not variable-dependent. 
I tested all of them. 

For the record, I spent way too much time over such a simple and not so useful thing, but that was a way to figure out how this app works. I learnt a lot of things on the way. :-)


JC5:
Nice, that's a pretty tedious job! I'll pick it up!

JC5:
Merged. Now let's hope the translators will understand the pipe character and apply it correctly.

PR #3417, opened 2020-06-01
Fix typo from #3413

sephrat:
Text was repeated twice in plural case (convert_expl_d_w + convert_select_destinations).


JC5:
:+1:

PR #3420, opened 2020-06-01
Display account names only once when displaying split transactions

### LINKED ISSUE #3409: Split transactions show source/destination accounts several times
See meta information block:
image
https://demo.firefly-iii.org/transactions/show/576

The account name should be displayed only once.

I'll try and fix this one too!

sephrat:
Fixes #3409
Not sure getAccounts is the best implementation - feel free to suggest how to improve it!


JC5:
Looks good!

PR #3424, opened 2020-06-03
Fix reconciliation difference calculation

sephrat:
Found an issue when a transaction is already (un)checked and you (un)check the "select all" button: the amount was added (subtracted) a second time to the balance.


JC5:
I can't merge because of a conflict. Could you pull the latest develop branch and recreate your PR? Sorry.

PR #3430, opened 2020-06-04
Autofill budget from last period

sephrat:
It's WIP but I wanted some feedback before fine-tuning it.

I've been working on this:
image
If a budget was set for the previous period, a new item appears in the dropdown list and allows you to set that amount.
image

Back-end looks solid to me. Front-end still needs some fine-tuning to handle all cases.
I'm open to other front-end suggestions.

Possible improvement would be to be able to do it in bulk - haven't thought of the UI yet (let the user decide which budget lines to import?).

### DIFF
--- a/app/Http/Controllers/Budget/IndexController.php
+++ b/app/Http/Controllers/Budget/IndexController.php
@@ -107,11 +107,15 @@ class IndexController extends Controller
         $spent           = '0';
         Log::debug(sprintf('1) Start is "%s", end is "%s"', $start->format('Y-m-d H:i:s'), $end->format('Y-m-d H:i:s')));
 
-        // new period stuff:
+        // new period stuff: 
         $periodTitle = app('navigation')->periodShow($start, $range);
         $prevLoop    = $this->getPreviousPeriods($start, $range);
         $nextLoop    = $this->getNextPeriods($start, $range);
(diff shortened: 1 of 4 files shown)


JC5:
It goes back per month, that could be it. 

But no, sorry.

sephrat:
I hoped it would not be too intrusive and robust enough. Seems like we won't find a middle ground here, and it's your project so I won't insist. 😃 

> I'd rather have that people use the auto budget method to add structure to their budgets.

I was really excited to use that but it just won't do for me. I thought it would initialise the amounts upon the first display, but the implementation through a cronjob is a huge drawback to me. I like to plan a few days ahead the coming period, I don't want to wait for the first day of the month to start planning.

JC5:
Don't get me wrong, I like the idea. It's a bit hacky but that's *not* your fault. This layout is at the end of it's time, we need something new. I'm already working on something else in another branch but that takes time.

sephrat:
Yeah I must say I didn't even know you could do budgets over multiple or overlapping period, the UI doesn't really give you a hint. Looking forward to a revamp!
I'd be happy to help if you need it. I'm not experienced enough in web tech to work on a complete module on my own but I can commit to making little QoL improvements like the ones I made lately.

PR #3455, opened 2020-06-09
Translate debug page

sephrat:
Only instructions are translated, debug info are left untouched so that we don't receive exotic bug reports. :-)


JC5:
Cool! :+1:

PR #3467, opened 2020-06-16
Fix #3437

### LINKED ISSUE #3437: Days left until payment in bill view
**Description**
I would like Firefly III to display the days left until bill payment for the bill not paid yet. For example in the bill view attached, instead of Not(yet), it could be added the days left for the payment. I think it is clearer in this way and could improve the visualization of the view knowing the days that you got for the bill coming.

**Solution**
Display the days left for a bill payment in the bill view. At least for the not yet paid or the not expected in this period values it would be useful. The already paid shows the most useful information.

**What are alternatives?**
Count the days until next expected match for knowing the days until next bill payment.

**Additional context**
imagen

sephrat:
Fix #3437
image

### DIFF
--- a/app/Http/Controllers/BillController.php
+++ b/app/Http/Controllers/BillController.php
@@ -212,6 +212,8 @@ class BillController extends Controller
         $bills = $unfiltered->map(
             function (Bill $bill) use ($transformer, $defaultCurrency) {
                 $return                            = $transformer->transform($bill);
+                $nextExpectedMatch = new Carbon($return['next_expected_match']);
+                $return['next_expected_match_diff'] = $nextExpectedMatch->isToday() ? trans('firefly.today') : $nextExpectedMatch->diffForHumans(today(), Carbon::DIFF_RELATIVE_TO_NOW);
                 $currency                          = $bill->transactionCurrency ?? $defaultCurrency;
                 $return['currency_id']             = $currency->id;
                 $return['currency_name']           = $currency->name;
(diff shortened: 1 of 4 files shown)


JC5:
- Try to avoid `today()`, or use it only when you're absolutely certain that the user's session range includes today. Since you're editing the tranformer, you must pull these values from the parameter bag. After all, you're now editing code that also works outside of the user session.
- How will your code respond when a user requests this data for a bill that was supposed to fire on Feb 26th, but didn't hit and  the date range (see the parameter bag!) is set to 1 May - 30 May? Will you leave it empty? Should it become "should have hit 1 month and 3 days ago? Or 2 months ago?
- When you edit the transformer, you must update the API documentation and describe exactly what can be expected in this field.

What you've written here is the happy flow, but there are many unhappy flows too. :wink:

sephrat:
Fair points. Thanks for the detailed review. Let me see what I can come up with.

sephrat:
Giving it a second thought, it probably wasn't the best idea to code this in the transformer. It doesn't make much sense to provide this data in the API, it can be calculated on the consumer side if needed. 
Passing a reference date in the parameterBag would be overkill IMO and I don't see a use case for this data when asking for past periods (your second point). Moreover, we would have to spaghetti the code to use DIFF_RELATIVE_TO_NOW or DIFF_RELATIVE_TO_OTHER for formatting. https://carbon.nesbot.com/docs/#api-humandiff 

I've moved the logic to the controller, is that better?

Issue #3476, opened 2020-06-21
Error connecting to database (i think)


revilo951:
I am running Firefly III version v5.2.8

**Description**
using docker-compose v3, fresh install of fireflyiii and postgres10. webpage error is:

```
Whoops! An error occurred.
Unfortunately, this error was not recoverable :(. Firefly III broke. The error is:

Trying to get property 'data' of non-object
```

**Extra info**
/debug page shows the same as the homepage, see above

I installed ping in the ffiii container, and i can ping the postgres container by hostname

logs attached below. LMK if there's something else you need. hopefully it's something I've fat-fingered, but if it is, i haven't caught it yet.

[ffiii-log.txt](https://github.com/firefly-iii/firefly-iii/files/4809080/ffiii-log.txt)

leander091:
Could you share the docker compose file used to install firefly-iii? 

It look like something is wrong with the authentication between your app and the database

JC5:
I see a “connection refused” in the logs.

revilo951:
my docker-compose.yml:

```
  fireflyiii:
    image: jc5x/firefly-iii:latest
    volumes:
      - /home/me/docker/firefly/export:/var/www/firefly-iii/storage/export
      - /home/me/docker/firefly/upload:/var/www/firefly-iii/storage/upload
    env_file: /home/me/docker/firefly/env/.env
    ports:
      - 85:80
    depends_on:
      - fireflyiiidb
    container_name: fireflyiii
    networks: 
      - firefly
  
  fireflyiiidb:
    image: postgres:10-alpine
    environment:
      - POSTGRES_USER=firefly
      - POSTGRES_PASSWORD=secret_firefly_password
    volumes: 
      - /home/me/docker/firefly/db:/var/lib/postgresql/data
    container_name: fireflyiiidb
    networks: 
      - firefly
    ports: 
      - 3306:3306
```


the .env db creds section:
```
# Database credentials. Make sure the database exists. I recommend a dedicated user for Firefly III
# For other database types, please see the FAQ: https://docs.firefly-iii.org/support/faq
# If you use Docker or similar, you can set these variables from a file by appending them with _FILE
# Use "mysql" for MySQL and MariaDB. Use "sqlite" for SQLite.
DB_CONNECTION=mysql
DB_HOST=fireflyiiidb
DB_PORT=3306
DB_DATABASE=firefly
DB_USERNAME=firefly
DB_PASSWORD=secret_firefly_password
```

[...]

PR #3479, opened 2020-06-21
Change default database in example environment file 

leander091:
Fixes issue #3476

Changes in this pull request:

- Switch the example environment file to Postgres to ensure consistency with the current docker image and documentation. 

It prevents error like: #3476

@JC5


JC5:
I'm migrating the default Firefly III configuration to MySQL. This is already the case in the `develop` branch but the main branch still refers to Postgres. The documentation also points to the main branch so people shouldn't find this file unless they look for it themselves.

leander091:
Alright, But i think this is still necessary for now.

The master branch has already the config for MySQL included in the example environment file.  See commit d493820cc82460749b9a531026edaed9e1caa2ce . So this breaks each new install when using docker compose.

JC5:
Crap. I thought I fixed that. I'm gonna push to master instead. Thanks for the headsup.

leander091:
No problem at all :)

PR #3506, opened 2020-07-01
Added transaction search api support

Johnny-Malizia:
- Transaction search api implemented using mostly the same process as the
existing search functionality.

@JC5

### DIFF
--- a/app/Api/V1/Controllers/Search/TransactionController.php
+++ b/app/Api/V1/Controllers/Search/TransactionController.php
@@ -25,7 +25,12 @@ declare(strict_types=1);
 namespace FireflyIII\Api\V1\Controllers\Search;
 
 use FireflyIII\Api\V1\Controllers\Controller;
+use FireflyIII\Support\Search\SearchInterface;
+use FireflyIII\Transformers\TransactionGroupTransformer;
+use Illuminate\Http\JsonResponse;
 use Illuminate\Http\Request;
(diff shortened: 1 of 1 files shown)


Johnny-Malizia:
Since the time I checked out the code, and started down the path of allowing transactions to be searchable via internal_reference and external_id, it seems support for searching by those values was added. I removed my changes in favor of the support you added, but it seems the query support utilizes a wild card like search.

I think that when trying to identify a transaction by an external identifier, it is best to use exact matches. For my specific use case, I want to lookup a transaction in firefly using the search api and know whether or not it exists. If it doesn't exist, I will create it, if it does exist it may be updated.

JC5:
Thanks for this excellent contribution! I'll take a good look later and merge it. :+1:

I would like to ask you to keep the search as is. The ability to search with a wildcard isn't something I'd like to change.

Johnny-Malizia:
> I would like to ask you to keep the search as is. The ability to search with a wildcard isn't something I'd like to change.

In regards to search specifically, are you open to a secondary way to search that allows exact matches instead of wildcard? For my use case I really need exact matching on these values but that doesn't necessarily need to be the only way the data can be queried.

JC5:
For the time being, no. I’ll think of something smart to make both of these options a possibility.

Johnny-Malizia:
Thank you. I've removed the changes to equality I made and I will enforce that on my client that performs the query.

PR #3573, opened 2020-07-17
Fix url

emansih:
`https://docs.firefly-iii.org/about-firefly-iii/features` is a 404 URL.

@JC5


JC5:
Nice, completely forgot that one.

Issue #3519, opened 2020-07-03
Locale not working (Windows 10) - possible workaround?


Crotha:
**Bug description**
In every Firefly-III version from 2.4.2 and up I encounter the same Problem:

"Invalid server configuration: Firefly III is unable to format monetary amounts because your server is missing the required packages. There are instructions how to do this. "

**Steps to reproduce**
Well, probably just using Windows

**Extra info**
So, I assume it is known already and from what I remember you said there is quite a problem with windows not using standard format locales or something like that.   
In any case, I wanted to make another Issue here, because the previous one rovided a (for now?) working solution: In #3306 @scottsousa described how he solved it for Windows 7.
After me stumbling and fumbling around, I finally found the string and commenting out the first two lines and adding in the one mentioned, it works for windows 10 too.

This is the change in the php-file:
![Change_Steam php](https://user-images.githubusercontent.com/47434645/86444775-97608400-bd11-11ea-9314-ab2b19f1e1cb.png)

I mainly wanted to mention it, because it seems like this is a windows-own problem that may be easily solved/bypassed in that way, if anyone else has this problem.
 
I couldn't find any other possible way to resolve the invalid server configuration otherwise, which is why I used version 2.4.1 or so until yesterday.
(I am using en-gb as main language, and en-ca as locale setting (because otherwise umlauts are shown wrongly, and en-ca has the preferred date-format))

I hope this wasn't to far off, and hope you have a nice day.   

Thanks for reading and also thanks a lot for firefly-iii :)

JC5:
You're very welcome. I'll write this down as an enhancement and pick it up in due course. This should work on Windows as well. Thanks for diving into this!

hoshsadiq:
@Crotha if you can try the latest dev branch, I reworked the way numbers are formatted. It would be good to get some feedback on it.

Crotha:
@hoshsadiq I downloaded the develop-branch, unpacked it and copied it over the old installation (of course, there are backups), I hope that is a correct way to do that...
FYI, just so you know, where my level of knowledge in these regards is.    

The "Invalid server configuration" thing is still there, BUT the numbers are formatted correctly (to US/UK standard).    
When setting the locale (not the language - language is still English(GB)) to German (Austria), the numbers formatting changes correctly everwhere but in one place: whenever you have to put in a number. 
E.g. budgeted amounts:     
[...]

Issue #3306, opened 2020-04-28
Locale [setlocale()] not working on Windows 7


scottsousa:
Hi James,

**Bug description**
I am running Firefly III version 5.2.4, and my problem is:

The locale functionality in the newer version of Firefly III is not working for me on Windows 7. I recently updated to version 5.2.4 from version 4.8.1.8 which did not have this issue. My locale is English (default; `en_US`).

Upon updating, I saw the following notice on the dashboard:

**Invalid server configuration:  Firefly III is unable to format monetary amounts because your server is missing the required packages. There are instructions how to do this.**

I know that Windows 7 is now EOL but I suspect this could be an issue on newer versions of Windows as well. After doing a bit of digging, it appears that Windows does not support UTF-8 locales. Here are some resources that I found (information on this topic appears to be scattered on the web and hard to find):

- [`setlocale()` PHP Manual (comment regarding Windows locale strings)](https://www.php.net/manual/en/function.setlocale.php#117529)
- [`setlocale()` Microsoft Docs](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/setlocale-wsetlocale?redirectedfrom=MSDN&view=vs-2019)
- [Windows Language Strings](https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/39cwe7zf(v=vs.100)?redirectedfrom=MSDN)
- [Stack Overflow Answer Regarding UTF-8 on Windows](https://stackoverflow.com/a/16120506)
- [Moodle Locale Docs (see "MS Windows" section)](https://docs.moodle.org/dev/Table_of_locales#MS_Windows)

It appears the issue within Firefly III is in the [`FireflyIII\Support\Steam::getLocaleArray()`](https://github.com/firefly-iii/firefly-iii/blob/e6442dd8affc6fed2e99c5ea9acdb90ab0af1465/app/Support/Steam.php#L613) function. I was able to "fix" this issue on my install by checking for Windows in the `PHP_OS` constant. I changed the return values in that function to "correct" (for Windows) values and the locale appears to be working.

For reference, this is the locale string format that worked for me:

```sprintf( '%s', str_replace( '_', '-', $locale ) )```

The return value is:

`en-US` (with a hyphen instead of an underscore and no `.UTF-8` or `.utf8` suffix)

I have not tested this with any other locales.

**Steps to reproduce**

1. Install (or upgrade) version 5.2.4 on Windows 7
[...]

Issue #3070, opened 2020-01-31
Currency symbol preceding amount regardless of locale


kbancerz:
**Bug description**
I am running Firefly III version 5.0.0, and my problem is:

Changing language and locale does not affect the position of the currency symbol as it should:

![image](https://user-images.githubusercontent.com/9020152/73561254-f8311b00-4458-11ea-9c02-5f19498ecb28.png)

Even though Polish is set, the symbol precedes the amount, rather than succeed it.

**Steps to reproduce**
Change the language and go to the cockpit.

**Extra info**
I am not too proficient in PHP, but I was able to find out that most probably the issue is related to these lines: https://github.com/firefly-iii/firefly-iii/blob/e9ab11051f95be8cd30c39d457f66e476233afcd/app/Support/Amount.php#L139-L140

Type-aware comparison is used (===), but the values assigned in these lines, are integers (0 or 1), not booleans: https://github.com/firefly-iii/firefly-iii/blob/e9ab11051f95be8cd30c39d457f66e476233afcd/app/Support/Amount.php#L137-L138

The comparison of boolean to integer always evaluates to false, so currency symbols always precede the amount.

If I modify the assignment lines, locales seem to work as expected:

```
$precedes = $amount < 0 ? $info['n_cs_precedes'] === 1 : $info['p_cs_precedes'] === 1;
$separated = $amount < 0 ? $info['n_sep_by_space'] === 1 : $info['p_sep_by_space'] === 1;
```

![image](https://user-images.githubusercontent.com/9020152/73562089-adb09e00-445a-11ea-813f-e6d219c755c7.png)

I don't know if this is relevant, but I use the official Docker version (apache-debian) as my running environment.

**Bonus points**

- [x] Nobody reported this bug before
- [ ] I have added a stack trace from my log files.
- [x] I have added a screenshot.
- [ ] I was able to replicate it on the demo site https://demo.firefly-iii.org/

JC5:
That's strange. The documentation is pretty clear:

https://www.php.net/localeconv

>  TRUE if currency_symbol precedes a positive value, FALSE if it succeeds one 

I'll dive into it. Thanks for debugging!

JC5:
I've pushed some fixes to the [demo site](https://demo.firefly-iii.org/), could you check it out?

kbancerz:
Pulled the fix on my local env and it works like a charm :) Thanks!

[...]

Issue #2946, opened 2020-01-03
Decimals are missing when formatting


Csardelacal:
**Bug description**
I am running Firefly III version 4.7.17.3, and my problem is:
That there are no decimals displayed across several places in the UI.
I checked the locales, they are properly installed, but the system won't work regardless.
Editing an amount displays a decimal separator, it's only in the reports and dashboard.

**Steps to reproduce**

Open the dashboard, the numbers are missing the decimal separators

**Extra info**

**Bonus points**
Debug information generated at 2020-01-03 14:09:57 Europe/Berlin for Firefly III version **4.7.17.3**.

| Variable | Content |
| --- | --- |
| FF version | 4.7.17.3 |
| FF API version | 0.9.2 |
| App environment | local |
| App debug mode | true |
| App cache driver | file |
| App logging | , daily |
| PHP version | 7.3.11-1\~deb10u1 |
| Display errors | Off |
| Session start | 2020-01-01 00:00:00 |
| Session end | 2020-12-31 23:59:59 |
| Session first | 2008-05-23 00:00:00 |
| Error reporting | ALL errors |
| Host | Linux |
| Interface | apache2handler |
| UserID | 1 |
| Attempt at "en" | false |
| Attempt at "English" | false |
| Attempt at "en_US.utf8" | 'en_US.utf8' |
| Attempt at "en_US.UTF-8" | 'en_US.UTF-8' |
| DB drivers | mysql |
| Current driver | mysql |
| Login provider |   |
| Storage disks | local-upload  |
| Using Sandstorm? | no |
| Is Sandstorm (.env) | false |
| Is Docker (.env) | false |
| bunq uses sandbox | false |
| Trusted proxies (.env) |  |
| User agent | Mozilla/5.0 (X11; Linux x86_64; rv:68.0) Gecko/20100101 Firefox/68.0 |
| Loaded extensions | Core, date, libxml, openssl, pcre, zlib, filter, hash, Reflection, SPL, session, sodium, standard, apache2handler, mysqlnd, PDO, xml, bcmath, calendar, ctype, curl, dom, mbstring, fileinfo, ftp, gd, gettext, iconv, imagick, intl, json, exif, mysqli, pdo_mysql, Phar, posix, readline, shmop, SimpleXML, soap, sockets, sysvmsg, sysvsem, sysvshm, tokenizer, wddx, xmlreader, xmlrpc, xmlwriter, xsl, zip, Zend OPcache |

[...]

PR #3578, opened 2020-07-18
Use php-intl to format currencies

hoshsadiq:
Currently the php function `number_format` is used to format currencies.
This is problematic as we have to figure out different things for
different currencies ourselves. These formats are determined based on
the libc's locale functions.

The issue arises where an OS doesn't have the proper locales installed,
or, in some cases, it's not supported (see below on multiple issues).

This addresses this issue by using the php-intl extensions to format the
numbers based on the locale. The extension is already a requirement in
`composer.json`. The solution does not rely on `LC_MONETARY` from the
underlying libc (which in Alpine Linux's case, which uses musl, is not
supported as of yet).

This isn't fully working (e.g. the balance on the dashboard doesn't work
properly yet), but I wanted to get your opinion on this first. It's been
a long time since I've done much PHP work so please give me any feedback
you have.

List of issues that are related and would potentially be fixed using
this PR:

- #2298
- #2946
- #3070
- #3306
- #3519

It would be great if you could tell me what different locales are to be
tested and what their expected input/outputs are. I'd be happy to write a
unit test beforehand, and ensure the PR conforms to previous code. I don't
want a whole bunch angry people coming to you telling you how the formatting
is suddenly messed up. I tested this with a couple of a languages, and they
*seem* okay.


JC5:
Thanks, this is very nice! And a compliment for fixing the tests as well, which are usually old and forgotten. Also by me. 😉 

A list of candidates can be found [in the configuration](https://github.com/firefly-iii/firefly-iii/blob/main/config/firefly.php#L329). 

Let me know how I can help!

hoshsadiq:
@JC5 this should be ready now. I tried my best to get the non-obvious currencies in the test list. Do let me know if there are any issues.

hoshsadiq:
By the way, I'm not really sure what to with the SonarCloud thing. The error doesn't seem to be related to my code changes.

JC5:
No worries. I started with Sonarcloud just recently and it's not yet working as I'd want to.

JC5:
Merged! Thanks a lot for your contribution :+1:

hoshsadiq:
@JC5 By the way, just in case this becomes a burden in the future, `NumberFormatter` uses [Non-breaking space](https://en.wikipedia.org/wiki/Non-breaking_space) pretty much everywhere. It took me a while to figure this out, so hoping you won't have to do the same in the future. I was thinking of doing a basic `str_replace` but figured that this character is actually better as it prevents a number from spanning to lines when it's at the end of the first line.

JC5:
Yeah that's quite alright. Breaking numbers would be pretty ugly. Thanks for the heads up!

JC5:
I committed some code to make sure negative numbers are displayed correctly. See my commit above this comment.

Issue #3382, opened 2020-05-22
Sign of transfer from savings account


McOrion:
**Bug description**
I am running Firefly III version 5.2.5. and my problem is: 

When making a transfer, initiated from the checking account on the dashboard, from the savings account to the checking account, the sign of the transfer is positive (green), both on the checkings account as on the savings account. I expected a positive (green) sign on the checking account, as the money is added there, but I expected a negative (red) sign on the savings account, as the money is withdrawn from this account.

**Steps to reproduce**
On the dashboard, create new transfer from the checking account. 

**Extra info**
It's also working like this on the demo website. See transaction "withdraw money from savings" of $ 256,-

**Bonus points**

- I searched and nobody reported this bug before---YES
- I have added a stack trace from my log files---NO
- I have added a screenshot.---YES
![Knipsel](https://user-images.githubusercontent.com/59847229/82701627-d3afb780-9c70-11ea-8403-a9669fd65e11.PNG)

- I was able to replicate it on the demo site https://demo.firefly-iii.org/ ---YES

JC5:
Indeed. This is on purpose. Transfers are blue and positive on both ends.

McOrion:
At least it works as designed ;-)

Looking close to my screen, I agree with you on the blue color. Although I'm not aware of having difficulties seeing colors, I didn't noticed the blue color earlier. That would be OK for me, as it makes clear the transaction is a transfer.

On my personal savings account I'm only able to transfer money in and out from my checkings account. So all transactions have the blue color and have a positive sign. That makes it difficult to see which transaction adds to the account and which do not. It would be great to have a sign, but a workaround would mean adding some keyword in the description.

JC5:
Sometimes you have to sacrifice stuff to solve it. I dumped the "grouped" view, making the code a lot simpler.

The view (.twig) is a lot more complex, granted, but this works fine. Fixed in the next release.

All credits to @sephrat for fixing the hard part!

PR #3603, opened 2020-07-27
Improvements to transactions list on home page

sephrat:
Changes in this pull request:
- Fixes issue #3382
- Started refactoring sign determination (still WIP - not very proud of the signature of signAmount)
- Fixed grouped transfers sum amount being displayed in black instead of blue

@JC5


sephrat:
Implementation is still not ideal but I believe it's a small step towards cleaner code.
My tests were quite thorough, hope I didn't miss anything:

- tested the home page and admin/links/show/x
- tested split and regular transactions for all kind of transaction, including reconciliation and initial balance

JC5:
Idk, on the whole I'm just not a fan. It can't escape the feeling it's spagetti code to drag the Account object back and forth like this. I have no better solution though.

sephrat:
Yeah I feel you. Trouble is the input format differs so much between all the methods. But it's not really better to copy/paste the same rules everywhere, there are already some rules that were missing here and there that should be fixed with this. 
Maybe the solution would be to fix the amount upstream, somewhere in Model/TransactionJournal. That logic doesn't really belong to Twig.

sephrat:
Oh wait, were you talking about the new Account object being passed from the template? For that I don't see a better solution either. This transfer thing is such an edge case that you need the context of the account.

JC5:
Yeah exactly. To be honest I’d rather drop the feature. But I’ll think about it. This particular list of transactions is pretty unique and over time pages like this one have become... less optimal :wink:

I have time this weekend I’ll mull it over.

sephrat:
Fair enough. I also have an idea for a different approach to this problem, but I'll do some trial and error before expanding my thoughts. :-) 
Should I withdraw the PR to dereference the open issue? 
Would you like me to submit another PR with just the refactoring part? Basically drop signAmountFromAccountPOV but keep signAmount?
There's also item #3 on my list that we can keep.

sephrat:
OK I've reverted the new feature but kept the refactoring part. It's still probably worth a major refactoring to allow new features, but I hope it's better than before.

PR #3638, opened 2020-08-02
UTF8 locale support for Windows

sephrat:
Saw #3519 and noticed it still didn't fix my Windows 10 dev system: special characters were displayed as �
With this change it's all good. Tested with French, Russian and Polish.
This is the result of my personal research but the condition comes from PHP documentation: https://www.php.net/manual/en/function.strftime.php example number 3.

@JC5

### DIFF
--- a/app/Support/Steam.php
+++ b/app/Support/Steam.php
@@ -614,6 +614,11 @@ class Steam
         if ('equal' === $locale) {
             return $this->getLanguage();
         }
+        
+        // Check for Windows to replace the locale correctly.
+        if (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN') {
+            $locale = str_replace('_', '-', $locale);
+        }
 
         return $locale;
     }
@@ -627,7 +632,6 @@ class Steam
         return [
             sprintf('%s.utf8', $locale),
             sprintf('%s.UTF-8', $locale),
-            str_replace('_', '-', $locale), // for Windows.
         ];
     }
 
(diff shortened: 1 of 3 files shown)


JC5:
Which characters are displayed badly, *exactly*? Is it in dates, translated strings or something else?

The PHP_OS thing is a hack i really dislike.

sephrat:
It is in dates:
image
image

Yeah I know it's dirty but it's even used by Carbon which is what I bumped into while debugging this.
image
(basically the snippet from PHP doc)

It's either this or we add the "-" variant on top of "_" to the array returned by getlocaleArray. I don't really mind.

sephrat:
Basically any UTF8 character will be displayed with �. In my previous example, it should write "août" instead of "ao�t".
Russian months are only made of � in current develop branch.

JC5:
Check, which means it's in the locales generated by PHP, not the translations.

JC5:
I'll take the array over the if-statement, at least my code will be cleaner for it.

sephrat:
Yeah the translations are fine.
No problem, go ahead.

JC5:
Thanks again!

sephrat:
Welcome. Ping me if you need a tester after your change.

sephrat:
Just noticed an issue in my fix: it doesn't work when locale is "equal" because of the return statement just before my code. We need to address that case too. Can I leave it up to you since you plan on reworking the condition?

JC5:
I'll pick it up, thx for the headsup!

PR #3640, opened 2020-08-02
Hide budgets list when it is empty

sephrat:
Before:
image

After:
image

Minor UX improvement for a new user IMO. Thought also about hiding the "Total available budget" frame, but I'll leave it up to you.

@JC5


JC5:
Let's keep that for now. Thanks!

PR #3641, opened 2020-08-02
Fix "Let's create a piggy bank!" not being displayed anymore

sephrat:
The recent implementation of groups to piggy banks does not handle correctly the case where no piggy banks are assigned to the default group.
When other piggy banks are assigned to a group, it is not visible. But when no piggy banks exist at all, the "Let's create a piggy bank!" frame is not displayed.
image

@JC5


JC5:
Elegant solution, merged!

PR #3696, opened 2020-08-13
Added revenue/liabilities *_account_inactive keys

psychowood:
en_US/firefly.php was missing

- revenue_account_inactive (used in /inactive-accounts/revenue)
- liabilities_account_inactive (used in inactive-accounts/liabilities)


JC5:
Nice. The other languages will pick this up, as you're undoubtedly aware.

psychowood:
> Nice. The other languages will pick this up, as you're undoubtedly aware.

I noticed, but only after creating another PR with both GB and IT translations, so I had to recreate the commit from scratch :)

Issue #3723, opened 2020-08-24
Support user search LDAP filter


glmdev:
**Description**
I self-host Firefly III and tie in authentication with an LDAP server I have running for SSO. This system works well, and I enjoy using Firefly III. However, I would like the ability to specify an LDAP filter for users in the environment config.

**Solution**
Adldap2 supports scopes, which can be added to the `config/ldap_auth.php` file. See the [docs here](https://adldap2.github.io/Adldap2-Laravel/#/auth/setup?id=scopes). It'd be nice to have a configurable scope that just applies whatever filter is set in the environment config, if one exists.

**What are alternatives?**
Not sure. There might be other ways to apply filters to Adldap2, but I'm unaware of them.

**Bonus points**
I'd need a bit of time to get an environment up and running, but I'd be willing to submit a PR for this, if it's something the maintainers are interested in implementing.

(P.S. thanks for the great application)

JC5:
You're very welcome 😁 

I'm not a fan. Keep in mind that this kind of complexity isn't really easily done using an environment variable. You'll have to stuff your config file full of if/else statements and I really dislike that. Because *everybody's* code gets slower for a feature very few people use. 

Can't you add this to the base DN?

glmdev:
That's a valid concern. The SSO solution I'm using, at least, filters application access using LDAP filters, so I'm afraid modifying the base DN isn't an option.

If the major part is clutter in the config files, my initial thought was to include the Adldap2 scope by default, and keep the logic to apply the filter or not w/in the scope. So, the only additions to the config files would be the single line for the scope, and the bit to fetch the ENV var. This doesn't avoid your unnecessary code point, however.

JC5:
That's a good idea!

JC5:
Since this is merged now, issue closed.

PR #3733, opened 2020-08-26
Add config option to specify custom LDAP filter

glmdev:
Fixes issue #3723

Changes in this pull request:

- Add ldap_auth.custom filter option
- Add ADLDAP_AUTH_FILTER env reference
- Create LdapFilterScope to apply the filter (if one exists) to LDAP queries

@JC5

### DIFF
--- a/app/Scopes/LdapFilterScope.php
+++ b/app/Scopes/LdapFilterScope.php
@@ -13,8 +13,8 @@ class LdapFilterScope implements ScopeInterface {
      */
     public function apply(Builder $query)
     {
-        $filter = config('ldap_auth.custom_filter');
-        if ( $filter ) {
+        $filter = (string) config('ldap_auth.custom_filter');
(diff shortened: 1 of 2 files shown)


JC5:
Nice. Two suggestions because I'm pedantic like that 😉 

1. Put `(string)` in front of the config call. That way, it's either "" or a valid filter, and never null or "" or a valid filter.
2. Check if it's not an empty string in your if statement, `'' !== $filter`.
3. If you're strict, you could also turn this into a single statement: `if(''!==(string)config('blabla'))`.

Thanks for the effort! 😁

glmdev:
Done :+1:

JC5:
Merged! :+1:

BWohlbrecht:
This feature is still missing from documentation.
Thanks for putting it in though.

JC5:
Feel free to edit the documentation

PR #3748, opened 2020-08-30
Wrong query params in autocomplete

psychowood:
Query string with two ?, broke some autocompletes. Probably an unlucky search&replace while migrating to the new API.

Btw, wouldn't be cleaner to use "expense" and "revenue" instead of "Expense account" and "Revenue account", since it is supported by the API?


JC5:
Thx!

Issue #3821, opened 2020-09-21
Add PWA Manifest


achembarpu:
Hi!

Thanks so much for Firefly! I've been using it without a hitch for months now. 😄 

I always use my (self-hosted) website on both mobile and desktop. For such websites, I tend to use Chrome's "Add shortcut -> Open as Window" feature to make it feel like an "App". I think it might generally benefit people like me, if there were a web app manifest - https://web.dev/add-manifest/.

Quick draft I put together: (to be served from `/manifest.webmanifest`)

```json
{
  "short_name": "Firefly",
  "name": "Firefly III",
  "description": "A personal finances manager",
  "icons": [
    {
      "src": "/favicon-32x32.png",
      "type": "image/png",
      "sizes": "32x32"
    },
    {
      "src": "/favicon-16x16.png",
      "type": "image/png",
      "sizes": "16x16"
    },
  ],
  "start_url": "/",
  "background_color": "<fill-me-in>",
  "display": "standalone",
  "scope": "/",
  "theme_color": "<fill-me-in>",
  "shortcuts": [
    {
      "name": "Asset Accounts",
      "short_name": "Assets",
      "description": "View asset accounts",
      "url": "/accounts/asset"
    },
    {
      "name": "Expense Transactions",
      "short_name": "Expenses",
      "description": "View expense transactions",
      "url": "/transactions/withdrawal"
    },
    ...
  ]
}
```

Do let me know what you think - I'm happy to help integrate this feature. Thanks!

JC5:
I already have two of those. [[1]](https://github.com/firefly-iii/firefly-iii/blob/main/public/manifest.json) and [[2]](https://github.com/firefly-iii/firefly-iii/blob/main/public/site.webmanifest). The second one is [referenced from the source](https://github.com/firefly-iii/firefly-iii/blob/main/resources/views/v1/partials/favicons.twig#L4) of Firefly III.

I guess one of them can be deleted?

achembarpu:
Aha, alright. I think the issue is that neither of them are named as per spec (https://www.w3.org/TR/appmanifest/).

Further, they should be served as so `<link rel="manifest" href="manifest.webmanifest">`.

JC5:
Thanks for looking it up. Feel free to submit a PR!

JC5:
Very cool! I merged the code. Your changes should be visible on the demo site already.

PR #3822, opened 2020-09-21
Fix web app manifest

achembarpu:
Hi!

As per discussion, this fixes issue #3821 . I have only updated v1's HTML, since I was unsure how to handle the v2 code.

Thanks!


JC5:
Nice work, merged :+1:

PR #3855, opened 2020-09-24
Add start_url to manifest file

achembarpu:
Hi,

Re #3821 , It appears `start_url` is a mandatory field. See Chrome DevTools output:

image

I'm serving Firefly over http, so you can ignore the first warning. The last warning is because Firefly is not offline-ready, i.e. does not use a service worker.

Thanks!


JC5:
Nice, merged. :+1:

Issue #3861, opened 2020-09-26
sonarqube analysis sonarcloud


lastlink:
**Description**
sonarqube gives good analysis of many language. sonarcloud is free for opensource. Would it be alright for me to do a github action that submits firefly for sonarqube analysis?

JC5:
Firefly III is already being analyzed by Sonarcloud.

JC5:
https://sonarcloud.io/dashboard?id=firefly-iii_firefly-iii

lastlink:
Could we sonar cloud badges or linked added to the readme then?

PR #3863, opened 2020-09-27
sonarcloud badges

lastlink:
* sonarcloud badges, easy fruits for open source contributors to work on. I can move the order around if preferred.

Fixes issue #3861 

Changes in this pull request:

- add sonarqube badges

@JC5

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -3,6 +3,16 @@
 [![Stargazers][stars-shield]][stars-url]
 [![Donate][donate-shield]][donate-uri]
 
+[![SonarCloud](https://sonarcloud.io/images/project_badges/sonarcloud-black.svg)](https://sonarcloud.io/dashboard?id=firefly-iii_firefly-iii)
+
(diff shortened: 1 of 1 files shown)


JC5:
Let's just do the main badge. You can replace the "stars" badge. Adding all of them is a bit much.

lastlink:
The main badge icon is bigger than your other badges so it'd have to go on its own line. Also even though you may not want the additional badges I'd recommend at least the line count badge which i think is useful to see when somebody opens the main page to this GitHub repo

JC5:
I see. That's a bummer. 

How about something else entirely? You could add a line to the "contributing" section of the file. Just skip the badges, and add something like "Firefly III is constantly scanned for bugs, issues and problems by sonarcloud. If you want to help improve ff3, check out the latest reports and take your pick!"

lastlink:
Ok I've tweaked it. you can preview it here https://github.com/lastlink/firefly-iii/tree/patch-3.

JC5:
Nice :+1:

PR #3957, opened 2020-10-18
Add base request that check user's auth

bu4ak:
Move duplicated code from all form requests to parent class.


JC5:
Thanks. It seems SonarCloud likes it. I guess some other static analyser will complain about tOo mAnY ExTeNsIoNs fRoM BaSe cLaSs next time.

Merged :+1:

PR #4015, opened 2020-10-31
Forward REMOTE_USER (remote_user_guard auth mode)

weimdall:
Changes in this pull request:
493d399 - Fix a typo

be421d0 - Fix an issue I had using "remote_user_guard", the REMOTE_USER header was not forwarded to the app -> Exception "The guard header was unexpectedly empty. See the logs."

I know my fix is not perfect : if someone edit AUTHENTICATION_GUARD_HEADER to something different .. then it won't work anymore. Maybe a note should be added, or do you have another idea to forward that header ?

@JC5

### DIFF
--- a/app/Support/Authentication/RemoteUserGuard.php
+++ b/app/Support/Authentication/RemoteUserGuard.php
@@ -67,9 +67,9 @@ class RemoteUserGuard implements Guard
 
             return;
         }
-        // Get the user identifier from $_SERVER
+        // Get the user identifier from $_SERVER or apache filtered headers
         $header = config('auth.guard_header', 'REMOTE_USER');
-        $userID = request()->server($header) ?? null;
+        $userID = request()->server($header) ?? apache_request_headers()[$header] ?? null;
         if (null === $userID) {
             Log::error(sprintf('No user in header "%s".', $header));
             throw new FireflyException('The guard header was unexpectedly empty. See the logs.');


JC5:
I think that's something for the documentation. I'll put it in.

As for the type, I'll pick it up through crowdin. I just noticed that particular string is currently untranslatable.

JC5:
Is there a reverse proxy between your auth solution and Firefly III, by any chance?

weimdall:
> Is there a reverse proxy between your auth solution and Firefly III, by any chance?

There is indeed, but that proxy is adding the header to the request 

I tested the flow :
- my client does the request to the reverse proxy on special firefly endpoint
- the reverse proxy add the header and forward the request to apache (inside the docker container)
- apache get the header (confirmed using log_forensic_module)
- Firefly app is called .. no more header

weimdall:
> As for the type, I'll pick it up through crowdin. I just noticed that particular string is currently untranslatable.

I removed 493d399 from the PR

weimdall:
By the way, if you or anyone is interested in a nginx reverse proxy conf on a specific endpoint that works with firefly, here is mine :

```
location /myendpoint/ {
    proxy_pass 				http://127.0.0.1:8080;
    proxy_set_header X-Forwarded-For 	$remote_addr;
    proxy_set_header X-Forwarded-Host 	$host;
    proxy_set_header X-Forwarded-Proto 	$scheme;
    proxy_set_header X-Forwarded-Port 	443;
    proxy_set_header Host 		$host;
    proxy_set_header X-Real-IP 		$remote_addr;
    proxy_set_header Accept-Encoding 	"";
    rewrite 	 ^/myendpoint/(.*)$ 	/$1 break;
    proxy_redirect   $scheme://$host 	$scheme://$host/myendpoint;
    sub_filter 	 $scheme://$host 	$scheme://$host/myendpoint;
    sub_filter 	 $scheme:\/\/$host 	$scheme:\/\/$host/myendpoint;
    sub_filter_types text/plain application/xml application/json text/html text/css text/javascript application/javascript;
    sub_filter_once  			off;
    sub_filter_last_modified 		on;
    gzip_types      text/plain application/xml application/json text/html text/css text/javascript application/javascript;
    gzip_proxied    no-cache no-store private expired auth;
}
```

JC5:
Thanks for the extra code. I'm going to add this to the documentation for sure. Like wise, I want to add your `.htaccess` addition but comment it out by default. Less is more. A short instruction maybe and a link back to the documentation.

weimdall:
I tried to optimize .htaccess to avoid any editing by user  ...
It has been a hard task, I read a lot of documentation and I came up with something like this :

```
RewriteCond %{ENV:AUTHENTICATION_GUARD} remote_user_guard
RewriteRule .* - [E=%{ENV:AUTHENTICATION_GUARD_HEADER}:%{HTTP:%{ENV:AUTHENTICATION_GUARD_HEADER}}]
```
But nested variable (%{HTTP:%{ENV:AUTHENTICATION_GUARD_HEADER}}) are not supported by RewriteRule ...

So, I went to Laravel documentation, and I found a specific function to access headers that apache filters out !
I edited the PR accordingly, that's definitely a better version (y)

It would maybe be interesting to switch the web server of the docker image to nginx, 
it's a lighter server, and this problem wouldn't happen ;)

JC5:
This will happen in nginx as well because headers with underscores in them are dropped. Using an Apache specific function will solve it in one case but not in all the others. Even then it's not guaranteed.

weimdall:
Well, at least this change make "remote_user_guard" mode working with the docker image and apache in general, since currently the header set in AUTHENTICATION_GUARD_HEADER is not even received by firefly.

I'll check what the behavior with nginx tomorrow.

weimdall:
With nginx, it is indeed required to add `underscores_in_headers on;` in order to get the REMOTE_USER, it will be then accessible as HTTP_REMOTE_USER in the classic request header ($_SERVER). Meaning that AUTHENTICATION_GUARD_HEADER will have to be set accordingly.

So, this PR will not fix all cases, but currently it is impossible to use the docker image set with "remote_user_guard" mode, I actually need to edit the image for that.
I am at least fixing that issue, and making things easier when using apache with a remote auth server.

JC5:
True, true. Merged/

PR #4050, opened 2020-11-09
Updated Transaction Search API to set limit from user preferences

Johnny-Malizia:
This change updates the transaction search api controller to use user preferences when setting a query limit as the previous config has been removed. Without this change, the search api always returns 0 entries.

@JC5


JC5:
:+1:

Issue #4060, opened 2020-11-14
Remote user guard: 500 error about type conversion


hakoerber:
**Bug description**
Firefly Version: 5.4.6

I am using the remote user authentication, which worked perfectly in 5.4.5. Since updating to 5.4.6, I'm getting the following error:

> Array to string conversion
> 
> This error occured in file `/var/www/html/app/Support/Authentication/RemoteUserGuard.php` on line 83 with code 0.

**Steps to reproduce**
* Use `AUTHENTICATION_GUARD=remote_user_guard`
* Use `AUTHENTICATION_GUARD_HEADER=HTTP_X_AUTH_USERNAME` (I am not sure whether this affects this bug, if you suspect it does I can also change this)
* Access firefly

**Some more Information**

The reported line was changed [this commit](https://github.com/firefly-iii/firefly-iii/commit/647c9c5eb027440e29c6a3b86f731640df9c0d36), so I suspect this is related to the fix for #4023. 

I'm not sure why the value of the username header should be an array. I checked, and it's set as a string in my authentication reverse proxy.

JC5:
Neither do I. Please add this line before the $emailAddress line:

```
Log::debug(sprintf('Header value: %s', var_export(request()->server($header), true)));
```

hakoerber:
it prints a bunch of headers, looks like *all* headers (redacted):

```
[2020-11-14 18:04:55] local.NOTICE: Header value: array (
  'HTTP_HOST' => '127.0.0.1:8080',
  'HTTP_CONNECTION' => 'close',
[...]
)
```

I debugged a bit further and found `$header` to be `NULL` in [this line](https://github.com/firefly-iii/firefly-iii/blob/bcb734a4593e1f368b9e6cc4f3157cbc667d6d9d/app/Support/Authentication/RemoteUserGuard.php#L82). This leads to leads to the dumping of **all** headers. It's of course empty because I did not set the `auth.guard_email` config (via env variable `AUTHENTICATION_GUARD_EMAIL`).

When I explictly specify the header via `AUTHENTICATION_GUARD_EMAIL`, all is well.

[...]

PR #4062, opened 2020-11-14
Fix remote guard email header

hakoerber:
Fixes issue #4060

Changes in this pull request:

- Do not fail with 500 on missing `AUTHENTICATION_GUARD_EMAIL` configuration

@JC5


hakoerber:
Is this SonarCloud check legit?

JC5:
Yes, but it doesn't just report on your code.

PR #4207, opened 2021-01-02
Fix 'spent per day' budget box

sephrat:
If no available budget is set, the budget box displays the total spent amount. The 'per day' amount is broken though, it's initialized to -1 but never set.
image

Tested with past and future periods as well.

@JC5

On a side note, I wonder if the negative amount is intended? If it's spent it's obviously negative. Same goes for 'Bills to pay' box.


JC5:
Nice :+1:

JC5:
Oh and expenses are always negative yes, even when you can deduce it from the context. I want to be consistent in how amounts are presented.

PR #4488, opened 2021-03-10
Set Japanese Yen decimal places to zero in the TransactionCurrencySeeder

KaihatsuOnline:
Changes in this pull request:

- Set Japanese Yen decimal places to zero.

@JC5


JC5:
:+1:

PR #4503, opened 2021-03-13
Fixed amountPerPeriod not respecting bill skips

Aethiran:
Changes in this pull request:
- Interface bills overview did not show expected monthy costs correctly (skips were ignored)
- amountPerPeriod() has been adapted to include skips

@JC5


JC5:
I like it, but this is a terrible way to cast to string: `''.($bill['skip'] + 1)`. Why not use `(string)`?

Aethiran:
Ah true, I will change chat

Aethiran:
*that

JC5:
Smart stuff. Merged!

PR #4627, opened 2021-04-09
added has_no_bill and has_any_bill search parameters

devfaz:
Changes in this pull request:

- adds search parameters "has_no_bill:true" and "has_any_bill:true"

@JC5


JC5:
If this is what I think it is, nice work! But you're not there yet.

- You must expand GroupCollectorInterface with abstracts of your new methods. Otherwise, it may not work correctly.
- You'll need to add translations ([en_US](https://github.com/firefly-iii/firefly-iii/blob/main/resources/lang/en_US/firefly.php#L451) is enough).

devfaz:
> If this is what I think it is, nice work! But you're not there yet.
> 
>     * You must expand GroupCollectorInterface with abstracts of your new methods. Otherwise, it may not work correctly.
> 
>     * You'll need to add translations ([en_US](https://github.com/firefly-iii/firefly-iii/blob/main/resources/lang/en_US/firefly.php#L451) is enough).

thx for the hints. Fixed :)

JC5:
Nice work! 🎉

PR #4716, opened 2021-04-26
Fix link in index

rubenvanerk:
Tiny fix for the index in the readme. Tried removing the space from the target header as well, but that breaks the image.

@JC5


JC5:
Nice, thanks :+1:

PR #4821, opened 2021-06-01
BUGFIX: sqlite does not return int types

JeroenED:
Fixes issue (none)

Changes in this pull request:
When trying to fin the reason why new budget limits were not created by the cronjob it appeared that sqlite does not return php integers for integer types. Instead it returns a string for everything.
This PR resolves this by casting the auto_budget_type to int

@JC5


JC5:
Nice, thanks!

Issue #4838, opened 2021-06-06
V2 UI caches API requests so that changes aren't visible in the UI


alex6480:
**Bug description**
API requests on the account index page are cached so that any changes aren't visible in the new UI. 

**Steps to reproduce**
Create an account and go to the account index.
Change something on the account (like renaming it) and go back to the index.
The change will not be visible because the page uses the cached API result from the last visit

If you clear the cache (Indexed DB) the changes will be visible upon refreshing the page

**Extra info**

**Bonus points**

- I searched and nobody reported this bug before

JC5:
This is all in progress, including a routine to reset the cache.

JC5:
Pushed some code to develop, the idea is that an added cache key will make sure the cache can be invalidated whenever necessary.

PR #4839, opened 2021-06-06
Add table to transaction index

alex6480:
Fixes issue # (if relevant)
Not relevant

Changes in this pull request:
I wanted to make some small UI changes, but then I realized the entire UI was being overhauled. So instead I decided to contribute to that.

I've implemented a table of transactions on the transaction index page similar to what was present in the old UI. I have not implemented the graphs or the monthly overviews, in order to keep this pull request small.
I've based the implementation on the code used for the account table in order to stay in line with your style of coding.

I've implemented functionality for split transactions and added a client side function to collapse them. They are implemented as separate rows in the table, but I've written a custom sorter that makes sure that they are displayed correctly. They could definitely be more clear in the UI, but I think that can be solved with some CSS. I just couldn't quite figure out what the best way to do it was, so I left it as it is.

It is affected by the bug I described in #4838 but I think a general solution is needed so I didn't attempt to solve it in this pull request.

billede

@JC5


JC5:
This works the same way as the account index, I presume? Because those tables are pretty slow with large datasets and most people have a lot of transactions. 

I can't check right now, but do me a favor: no sort options in this table, and make sure each page is loaded on-demand (no preloading).

JC5:
Yes, saw the code, and the result. I like it!

Will be merged into develop. Good work!

alex6480:
The code currently does preload the entire table and has client side table.
I was in the middle of fixing it, when I got side tracked by the fact that this line in the Transaction/ShowController is incredibly slow 

    $manager->createData($resource)->toArray())

It takes several seconds for it to process an API request, which makes no sense as it takes less than a second to actually run the query on the database. I'm not that well-versed in Fractal so I'm not quite sure what's taking so long.

JC5:
I just pushed a basic fix, sorry for getting in your way. Feel free to do it better 😁 

That code is terrible, so no surprise there. The collection is good, query is too, but each object is being transformed individually, and that takes ages. Let me pick up that one :+1:

alex6480:
I figured out what's going on.

Laravel lazy loads missing relations. 
So when the toArray function is called, the following queries will run once for each transaction in the set.

    select `tags`.`tag` from `tag_transaction_journal` left join `tags` on `tag_transaction_journal`.`tag_id` = `tags`.`id` where `tag_transaction_journal`.`transaction_journal_id` = ? order by `tags`.`tag` asc
    select `name`, `data` from `journal_meta` where `transaction_journal_id` = ? and `name` in (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?) and `deleted_at` is null,"sepa_cc","sepa_ct_op","sepa_ct_id","sepa_db","sepa_country","sepa_ep","sepa_ci","sepa_batch_id","internal_reference","bunq_payment_id","import_hash_v2","recurrence_id","external_id","original_source","external_uri","recurrence_count","recurrence_total"] 
    select `name`, `data` from `journal_meta` where `transaction_journal_id` = ? and `name` in (?, ?, ?, ?, ?, ?) and `deleted_at` is null 
    select * from `transaction_journals` where `transaction_journals`.`user_id` = ? and `transaction_journals`.`user_id` is not null and `transaction_journals`.`id` = ? and `transaction_journals`.`deleted_at` is null limit 1
    select * from `locations` where `locations`.`locatable_id` = ? and `locations`.`locatable_id` is not null and `locations`.`locatable_type` = ? limit 1
    select * from `notes` where `noteable_id` = ? and `noteable_type` = ? and `notes`.`deleted_at` is null limit 1

JC5:
Thanks for debugging, i'll pick it up :+1:

PR #4855, opened 2021-06-16
Update api.php

vonsogt:
to avoid getting error when clear route cache:
`Unable to prepare route [api/v1/insight/transfer/asset] for serialization. Another route has already been assigned name [api.v1.insight.income.asset].`

@JC5


JC5:
Nice find, never noticed!

Issue #4935, opened 2021-07-23
Images in the readme are broken


Flightkick:
**Bug description**
Images in the readme refer to `https://www.firefly-iii.org/static/screenshots/` these resources do not seem to be served anymore. Which has caused the images to break.

**Steps to reproduce**
Check the readme at: https://github.com/firefly-iii/firefly-iii

**Extra info**
I've prepared a PR with a proposed solution. This issue is for reference/tracking.

JC5:
Fixed :+1:

PR #4936, opened 2021-07-23
Feature/fix broken images readme

Flightkick:
**Fixes issue # (if relevant)**
https://github.com/firefly-iii/firefly-iii/issues/4935

**Changes in this pull request:**
The first thing I've checked was if the resources had been moved and exist somewhere in https://github.com/firefly-iii/firefly-iii.org/tree/main/src
This did not seem to be the case. I came up with a few solutions:
1. Add images to the firefly-iii.org repository
2. Add the images to the readme by pasting it in the online Github editor.
3. Add the images to the firefly-iii repository

The first option may be the easiest, if you prefer that you can always close this PR in favor of adding the images to that website. I didn't pick this option because the website doesn't use the images and you'd lose the advantages of Github's CDN.

The second option would've been quick as well but that would probably have tied the images to my account and I'm not sure if something will break if Github ever decides to expire images from user-content.githubusercontent.com

So I chose the third option of adding the images to the `.github` directory of this project. I've updated the readme itself as well to refer to those images but this won't work until all changes have been merged into main (as the readme images now refer to `https://raw.githubusercontent.com/firefly-iii/firefly-iii/main/.github/assets/img/`)

Also moved the iPhone image out of the header.

@JC5


JC5:
Looks good, great solution!

PR #5007, opened 2021-08-21
Add additional extensions to composer require

hoshsadiq:
Not sure if this all is correct, however I struggled to get it running without these additional extensions. Composer didn't complain but I ended up getting plenty of errors around these. I don't know if any of them are optional or not but happy to move them to the suggests section if you need me to.


JC5:
I don't mind merging this, but with the exception of mbstring these are all part of the default PHP installation. Unless explicitly removed. Not sure what to do here.

JC5:
Oh well, doesn't hurt.

hoshsadiq:
Thanks! For me it was using Alpine Linux that caused the errors so I figure it wouldn't hurt to add them.

Issue #5064, opened 2021-09-08
Maskable icon and service worker


LBreda:
### Support guidelines

- [X] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)
- [X] My request is not listed as [an idea that won't be implemented](https://docs.firefly-iii.org/firefly-iii/support/features/#ideas-that-wont-be-implemented)
- [X] I've used [the search](https://github.com/firefly-iii/firefly-iii/issues?q=is%3Aissue) and this has not been discussed before.

### Description

Firefly III does not have a maskable icon (a icon that looks good on most  mobile devices) and a service worker (a js script that provides basic services to let Android, Windows and iOS devices to install it like an app).

It would be a nice functionality and a very easy one to implement.

### Solution

It would be nice to add a set of maskable icons and a service worker.

### Additional context

I could do the task, but I'd like to know if it is considered a nice thing to have by the developer. I would need a SVG version of the logo.

JC5:
Feel free to submit a PR! You can find the logo in [the documentation](https://docs.firefly-iii.org/firefly-iii/about-firefly-iii/logo/)

Are you sure the correct icon isn't already [in place](https://github.com/firefly-iii/firefly-iii/tree/main/public)? I used the real favicon generator to build it.

LBreda:
The current favicon is a good favicon but the mobile systems usually cut the icon in different shapes by masking it and need a maskable icon.

You can verify how the current icon fails to adapt in different shapes by using [maskable.app](https://maskable.app/), and you can find details on (web.dev)[https://web.dev/maskable-icon/].

I'll make a PR!

LBreda:
![immagine](https://user-images.githubusercontent.com/633783/132588962-cae84803-1be7-4bcc-993d-df89667da29b.png)

Is something like this OK or I should use the pig tail?

LBreda:
![immagine](https://user-images.githubusercontent.com/633783/132589417-ed3877b4-1a34-44df-93e5-54253d2962c5.png)

Is something like this OK or should I use the tail only?

JC5:
Looks good!

LBreda:
I need to add a new offline page. I tried to add a new (publicly available) route:

```php
Route::group(
    ['middleware' => 'binders-only'],
    static function () {
        Route::get('offline', fn () => view('errors.Offline'));
    }
);
```

but Laravel ignores it, even clearing the cache. What am I missing?

[...]

PR #5067, opened 2021-09-11
Maskable icon and service worker

LBreda:
Solves #5064 

Changes in this pull request:

- Added a set of maskable icons which look good on mobile devices
- Made the web app a installable PWA via a service worker

I tried to add the service worker in the Laravel Mix manifest but breaks it an I couldn't find why. The cache stops working, and it is required for a installable PWA.

@JC5


JC5:
Looking good! Thanks!

A few comments, then it's ready to merge I think.

- Please remove .idea from the gitignore file. It's expected to add IDE specific files to your own gitignore_global file.
- The copyright of serviceworker belongs to you, you can refer to your github profile or email address
- Drop the translations, *expect* en_US. I will auto-generate the necessary lines using [Crowdin](https://crowdin.com/project/firefly-iii)
- Rename "Offline.twig" to "offline.twig" to be consistent with other resources, and update the route to match

Also very happy to see you touched and improved the other icons as well. Nice work! :+1:

JC5:
Nice work, merged!

PR #5153, opened 2021-10-02
healthcheck endpoint

ajgon:
Changes in this pull request:

- This PR adds a /health endpoint which does nothing more than returning plaintext OK response. It is useful for service discovery tools (consul/kubernetes etc.), to quickly say, that service is alive and running. 

@JC5

### DIFF
--- a/app/Http/Controllers/HealthcheckController.php
+++ b/app/Http/Controllers/HealthcheckController.php
@@ -1,0 +1,43 @@
+<?php
+/**
+ * HealthcheckController.php
+ * Copyright (c) 2019 james@firefly-iii.org
+ *
+ * This file is part of Firefly III (https://github.com/firefly-iii).
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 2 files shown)


JC5:
You mean, apart from pinging any other page or static resource?

JC5:
Doing some Googling on this makes me inclined to have something in line with https://inadarei.github.io/rfc-healthcheck/

ajgon:
Yes :D 

Doesn't static resources are handled by webserver (nginx/apache) directly, not the app itself? Meaning, that if app is dead (for whatever reason), static resources would still work?

Also static resources are still larger responses than `OK` :)

> Doing some Googling on this makes me inclined to have something in line with https://inadarei.github.io/rfc-healthcheck/

This is interesting, I wasn't aware of that - but isn't a bit overkill? Most of the service discovery tools (that I'm aware of), usually doesn't care about response as long as it's 200.

JC5:
> This is interesting, I wasn't aware of that - but isn't a bit overkill? Most of the service discovery tools (that I'm aware of), usually doesn't care about response as long as it's 200.

Idk but i'd rather follow a standard than invent the next one. Even if it's just static JSON (for now).

ajgon:
> > This is interesting, I wasn't aware of that - but isn't a bit overkill? Most of the service discovery tools (that I'm aware of), usually doesn't care about response as long as it's 200.
> 
> Idk but i'd rather follow a standard than invent the next one. Even if it's just static JSON (for now).

I think it's a draft for now, but it's less important. The thing which I don't like in this draft, is the large response. Service discovery tools, pings containers/vms pretty often - and with this, each time they will be fetching a lot of unnecessary data. 

Also, from my personal experience, I implemented healthchecks in many services, and this is the first time I see this. So if it's a standard - it's not well adopted one :)

JC5:
Let me validate & merge, stay tuned 👍

Issue #5110, opened 2021-09-23
JavaScript page/text after storing transaction (iOS / Firefox Mobile) - Firefly III 5.6.0 - Layout v2


JanFeld:
### Support guidelines

- [X] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)

### I've found a bug and checked that ...

- [X] ... the documentation does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem

### Description

When storing a transaction on my mobile phone (iOS 15 / Firefox) i'll get forwarded to a text page which looks like the in the images attached.

The transaction is stored and seems to be correct.
I can access all pages afterwards without any issues.
The "Transaction .. stored" message is displayed on the first access of any FF3 page afterwards.

### Expected behaviour

Simply go back to the last page where i've been coming from.
Including a message that the transaction has been stored.

### Debug information

* iOS 15.0
* Firefox Daylight 37.0 (5971)
* Firefly 3 5.6.0 - Layout v2

### Steps to reproduce

Use a mobile phone (preferably iOS with OS 15 and Firefox) and try to store a transaction using the layout v2.

### Additional info

Sadly i'm not able to reproduce this issue on the demo page because i've forgot how to get to the layout-v2 preview..

The screenshots below show the page.

![ff3_bug_01](https://user-images.githubusercontent.com/22739965/134554319-ac1d30a6-cc03-4377-8c02-a2256865cb3d.png)

![ff3_bug_02](https://user-images.githubusercontent.com/22739965/134554327-054159ae-88b2-4a26-9aa0-14c7a52ce362.png)

Cheers,
Jan

JC5:
Thanks, I noticed it as well. I hoped it was a glitch. I'll pick it up!

schneekluth:
It also appears with the legacy UI on iOS 15.

JC5:
Fixed in the next release

rygos:
i have the same Problem. serviceworker.js is viewed every time I submit a transaction (new or edit).
Safari on macOS Monterey. On Chrome everything works as expected.

JC5:
Which version of Firefly III are you running, please?

rygos:
The latest. 5.6.1.
I can reproduce the error with the installed version on my server or also on your demo installation.

When I add or edit a transaction, after clicking on the Save button, this Javascript page appears, which is shown in the screenshots above.

The URL of showed page (after edit):
[...]

PR #5156, opened 2021-10-02
Bugfix: Safari redirecting to serviceworker.js

z1r0-:
Fixes issue #5110

Changes in this pull request:

The fix in storeCurrentUrl was good, but the URL used for the redirect was taken from `app('url')->previous()` which seems to have another source? (haven't used laravel for a very long time) However `session()->previousUrl()` contains the correct saved url from storeCurrentUrl and fixes the issue with redirect to serviceworker.js in safari on iOS and macOS.

@JC5

### DIFF
--- a/app/Support/Http/Controllers/UserNavigation.php
+++ b/app/Support/Http/Controllers/UserNavigation.php
@@ -172,7 +172,7 @@ trait UserNavigation
      */
     final protected function rememberPreviousUri(string $identifier): ?string
     {
-        $return = app('url')->previous();
+        $return = session()->previousUrl();
         /** @var ViewErrorBag|null $errors */
         $errors    = session()->get('errors');
         $forbidden = ['json', 'debug'];


JC5:
I went into a slightly different direction, but credit where credit is due.

The `session()->previousUrl()` contains the right url because it is configured to filter the session's previous url. This is basically what you need anyway so your solution is correct.

What I did was, centralise the URL cleaning methods (there were 3) in Steam.php and use that one in `session()->previousUrl()`. Then, add `app('steam')->getSafePreviousUrl()` that refers back to the session method (your suggestion).

In the code, all usages of `url()->previous()` and `URL::previous` are dropped.

- https://github.com/firefly-iii/firefly-iii/blob/develop/app/Support/Steam.php#L52
- https://github.com/firefly-iii/firefly-iii/blob/develop/app/Support/Steam.php#L72

Issue #5276, opened 2021-11-13
The GET method is not supported for this route. Supported methods: POST


GaneshKandu:
### Support guidelines

- [X] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)

### I've found a bug and checked that ...

- [X] ... the documentation does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem

### Description

I am getting this issue while trying to logout.

MethodNotAllowedHttpException
HTTP 405 Method Not Allowed
The GET method is not supported for this route. Supported methods: POST.

### Expected behaviour

it should logout successfully 

### Debug information

Debug information generated at 2021-11-13 10:14:47 Europe/Amsterdam for Firefly III version **5.6.3**.

| Scope | Version |
| --- | --- |
| Firefly III | 5.6.3 |
| Firefly III API | 1.5.4 |
| PHP | 8.0.7 |
| Host | Linux |

| System info | Value |
| --- | --- |
| Installation ID | 9ec32e13-4cac-41ea-9dae-c1c445c64663 |
| Using docker? | false |
| Telemetry | false |
| Layout | v1 |
| System TZ | Europe/Amsterdam |
| Browser TZ | Asia/Kolkata |
| App environment | local |
| App debug mode | true |
| App cache driver | file |
| App logging | notice, daily |
| Display errors | Off |
| Error reporting | ALL errors |
| Interface | apache2handler |
| Default language | en_US |
| Default locale | equal |
| BCscale | 24 |
| DB drivers | mysql, sqlite |
| Current driver | mysql |
| DB version | 18 (exp. 18) |
| Login provider | eloquent  |
| Trusted proxies (.env) |  |

| User info | Value |
| --- | --- |
| Session start | 2021-11-01 00:00:00 |
| Session end | 2021-11-30 23:59:59 |
| Session first | 2021-01-01 00:00:00 |
| User ID | 1 |
| User language | en_US |
| User locale | en_US |
| Attempt at "en_US.utf8" | 'en_US.utf8' |
| Attempt at "en_US.UTF-8" | 'en_US.UTF-8' |
| User agent | Mozilla/5.0 (Windows NT 6.3; Win64; x64; rv:94.0) Gecko/20100101 Firefox/94.0 |

### Steps to reproduce

Install Firefly 5.6.3 and login
then click on logout button

JC5:
[...]

PR #5277, opened 2021-11-13
ERROR : The GET method is not supported for this route. Supported met…

GaneshKandu:
…hods: POST.

#5276 

This works for me.

Fixes issue # (if relevant)

Changes in this pull request:

- 
- 
- 

@JC5


JC5:
Sorry, logout will not be a GET route due to CSRF dangers.

Issue #5231, opened 2021-11-02
Allow setting LDAP model (FreeIPA/OpenLDAP/ActiveDirectory...)


StillLoading:
### Support guidelines

- [X] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)

### I've found a bug and checked that ...

- [X] ... the documentation does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem

### Description

I just finished updating to the latest version (5.6.2) and updated my LDAP configuration. It seems like the default is set to use the OpenLDAP schema which fails with FreeIPA since there is no '[entryuuid]' attribute (see logs).

I found a workaround by editing `config/auth.php` and `app/Ldap/AttributeHandler.php` to use `LdapRecord\Models\FreeIPA\User::class` instead of `LdapRecord\Models\OpenLDAP\User::class`.

Since the old `ADLDAP_CONNECTION_SCHEME=` variable seems to be deprecated it would be nice to have something similar with the new LDAP library.

### Expected behaviour

Allow admins to set the LDAP schema in `.env`.

### Debug information

`
[2021-11-02 16:26:50] local.INFO: LDAP (ldap://**removed**:389) - Operation: Search - Base DN: **removed** - Filter: (&(objectclass=\74\6f\70)(objectclass=\70\65\72\73\6f\6e)(objectclass=\6f\72\67\61\6e\69\7a\61\74\69\6f\6e\61\6c\70\65\72\73\6f\6e)(objectclass=\69\6e\65\74\6f\72\67\70\65\72\73\6f\6e)(uid=\6a\73)) - Selected: (entryuuid,*) - Time Elapsed: 34.8
[2021-11-02 16:26:50] local.INFO: User [**removed**] has been successfully discovered for authentication.
[...]

PR #5279, opened 2021-11-13
Develop

StillLoading:
Saw the (partial) fix for #5231 and noticed it did not include FreeIPA. So I fumbled this together.

This is just a suggestion on how to handle the available models. You don't have to use it as is, I can change it if it doesn't fit your code style/scheme.

### DIFF
--- a/config/auth.php
+++ b/config/auth.php
@@ -24,8 +24,21 @@ declare(strict_types=1);
 use FireflyIII\Ldap\AttributeHandler;
 use FireflyIII\Ldap\Rules\UserDefinedRule;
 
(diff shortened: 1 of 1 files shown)


JC5:
Good idea! :+1:

JC5:
Only thing I realise (after merging) is that you can drop the OpenLDAP check; it's the default anyway.

StillLoading:
> Only thing I realise (after merging) is that you can drop the OpenLDAP check; it's the default anyway.

Sorry, that was my "better safe than sorry" approach since I'm not familiar with the code base. I'll whip up another PR.

PR #5371, opened 2021-12-07
Bugfix: Empty pre-filled value in auto-budgets

weimdall:
The auto-budget pre-filled value is empty when the set amount is above 1000 (view /budgets/edit/X).

The root cause was the function 'number_format' returning a string that
cannot be casted to a float when the amount is above 1000, because of the
thousands separator.

Introduced by commit b2eeeed0af in 5.6.5.

@JC5


JC5:
:+1:

JC5:
Nice, merged!

Issue #5606, opened 2022-01-23
Search operators for `external_url` field


GeorgeHahn:
### Support guidelines

- [X] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)
- [X] My request is not listed as [a very good idea, but unfortunately...](https://docs.firefly-iii.org/firefly-iii/about-firefly-iii/what-its-not/)
- [X] I've used [the search](https://github.com/firefly-iii/firefly-iii/issues?q=is%3Aissue) and this has not been discussed before.

### Description

It would be helpful to be able to search for transactions by `external_url`. In particular, I'd like to be able to filter for transactions with a blank `external_url` as part of an import tool.

### Solution

A search operator similar to the `no_notes:true` operator would be very helpful.

JC5:
- https://github.com/firefly-iii/firefly-iii/issues/5326
- https://github.com/firefly-iii/firefly-iii/issues/4903

Issue #5326, opened 2021-11-28
Automation Trigger: Internal Reference contains..., begins with...., ends with...


jasonwinget:
### Support guidelines

- [X] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)
- [X] My request is not listed as [a very good idea, but unfortunately...](https://docs.firefly-iii.org/firefly-iii/about-firefly-iii/what-its-not/)
- [X] I've used [the search](https://github.com/firefly-iii/firefly-iii/issues?q=is%3Aissue) and this has not been discussed before.

### Description

Automation triggers for "Internal Reference" only has a "is..." trigger.  

### Solution

Please expand with contains, begins with, and ends with to complement the other triggers.

### Additional context

![image](https://user-images.githubusercontent.com/3647696/143725304-ef4b2deb-601b-4226-9cb0-8ac07d8d3327.png)

JC5:
I'll write it down!

Issue #4903, opened 2021-07-08
Add rule triggers for optional fields


pwsoftwaredesigns:
**Description**
It would be useful to have rule triggers available for optional fields.  For example, I wanted to add a rule for the "Processing Date" field if it is empty, but could find no available triggers for this.

JC5:
Good idea, it's on the list.

pwsoftwaredesigns:
I wanted to add something to this suggestion and will use "Processing Date" as an example.  I use the "Date" field to store when I actually made the purchase and the "Processing Date" field to store the date that it posted within my bank statement. Usually, the processing date is two days after the purchase date unless that happens to fall on a weekend, in which case the processing date will be the following Monday.  I am wondering if the trigger rules allow (or can be made to allow) for math/script operations to be performed to automatically set the field based upon the aforementioned conditions.

PR #5631, opened 2022-01-25
Attempt to add `external_url` filtering support

GeorgeHahn:
Related to https://github.com/firefly-iii/firefly-iii/issues/5606 / https://github.com/firefly-iii/firefly-iii/issues/5326 / https://github.com/firefly-iii/firefly-iii/issues/4903

Changes in this pull request:

- Adds rules/search support for presence/absence of the `external_url` field.

Todo:
- [ ] Set up a local dev environment and test this out.

@JC5 Are you open to this change?

### DIFF
--- a/app/Helpers/Collector/Extensions/MetaCollection.php
+++ b/app/Helpers/Collector/Extensions/MetaCollection.php
@@ -215,6 +215,36 @@ trait MetaCollection
     /**
      * @inheritDoc
      */
+    public function withoutExternalUrl(): GroupCollectorInterface
+    {
+        if (false === $this->hasJoinedMetaTables) {
+            $this->hasJoinedMetaTables = true;
+            $this->query->leftJoin('journal_meta', 'transaction_journals.id', '=', 'journal_meta.transaction_journal_id');
+        }
+        $this->query->where('journal_meta.name', '=', 'external_url');
+        $this->query->whereNull('journal_meta.data');
+
+        return $this;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function withExternalUrl(): GroupCollectorInterface
+    {
+        if (false === $this->hasJoinedMetaTables) {
+            $this->hasJoinedMetaTables = true;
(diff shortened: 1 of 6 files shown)


JC5:
Nice. I'd say "transaction has any external URL". Otherwise, seems to look good already. I'll pull the PR later and check it out.

JC5:
In its current form, it seems to skip over any transaction that doesn't have the external_url meta data value.

The `no_external_url:true` operator finds nothing, because there are no entries where the name is external_url but the value is empty. Likewise, if they do have the external_url, its never empty.

I solved it with this query. I'm not sure if i can push to your branch.

```php
    /**
     * @inheritDoc
     */
    public function withoutExternalUrl(): GroupCollectorInterface
    {
        if (false === $this->hasJoinedMetaTables) {
            $this->hasJoinedMetaTables = true;
            $this->query->leftJoin('journal_meta', 'transaction_journals.id', '=', 'journal_meta.transaction_journal_id');
        }
        $this->query->where(function(Builder $q1) {
            $q1->where(function(Builder $q2) {
                $q2->where('journal_meta.name', '=', 'external_url');
                $q2->whereNull('journal_meta.data');
            })->orWhereNull('journal_meta.name');
        });


        return $this;
    }
```

The `any_external_url` filter is working as expected.

Since the search uses the exact same operators, you can easily test it over there. To that end, add two translations to `firefly.php` (you know which one):

```php
    'search_modifier_no_external_url'                     => 'The transaction has no external URL',
    'search_modifier_any_external_url'                    => 'The transaction must have a (any) external URL',
```

I'm not sure if the grammar rules make it "a" or "an" in this case.

GeorgeHahn:
@JC5 I was worried about that query; thanks for the fix. I pushed your solution & the search strings.

I'm having some trouble getting the app running in a devcontainer. Are you happy with the testing you did or should I spend some time to test this out another day?

JC5:
It looks good to me, I'll merge later today and pick up the translations etc.

JC5:
Merged :+1:

Issue #5847, opened 2022-03-03
Unauthenticated (401) on /transactions with remote auth


weimdall:
### Support guidelines

- [X] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)

### I've found a bug and checked that ...

- [X] ... [the documentation](https://docs.firefly-iii.org/) does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem
- [X] ... it's [definitely me, not you](https://github.com/firefly-iii/firefly-iii/blob/main/.github/its_you_not_me.md)

### Description

-- This only happen while using remote guard headers to authenticate user --
When I try to edit / create a transaction, the resulting xhr request (`POST /api/v1/transactions?_token=XxXxXxXxXxX`) get the error code `401 Unauthorized` with the following body :
`{"message":"Unauthenticated","exception":"AuthenticationException"}`

### Expected behaviour

I expected to be able to edit / create / delete transactions as in 5.6.14

### Debug information

Debug information generated at 2022-03-03 17:40:54 America/XX for Firefly III version **5.6.16**.

| Scope | Version |
| --- | --- |
| Firefly III | 5.6.16 |
| Firefly III API | 1.5.5 |
| Build | 113, 02-03-2022 19:39:22 CET |
| PHP | 8.0.16 |
| Host | Linux |

| System info | Value |
| --- | --- |
| Using docker? | true |
| Layout | v1 |
| System TZ | America/XX |
| Browser TZ | America/XX |
| App environment | production |
| App debug mode | false |
| App cache driver | file |
| App logging | notice, stack |
| Display errors | Off |
| Error reporting | ALL errors |
| Interface | apache2handler |
| Default language | en_US |
| Default locale | equal |
| BCscale | 24 |
| DB drivers | mysql, pgsql, sqlite |
| Current driver | mysql |
| DB version | 18 (exp. 18) |
| Login provider | eloquent  |
| Trusted proxies (.env) | ** |

| User info | Value |
| --- | --- |
| Session start | 2022-03-01 00:00:00 |
| Session end | 2022-03-31 23:59:59 |
| Session first | 2018-08-16 00:00:00 |
| User ID | 1 |
| User language | en_US |
| User locale | en_US |
| Attempt at "en_US.utf8" | 'en_US.utf8' |
| Attempt at "en_US.UTF-8" | 'en_US.UTF-8' |

### Steps to reproduce

[...]

PR #5863, opened 2022-03-08
Bugfix #5847

weimdall:
Fixes issue #5847
On my side, by commenting this line, the bug is fixed.
Though, this is probably not a suitable solution.

@JC5


JC5:
No it's not.

PR #6101, opened 2022-05-24
feat: add the ability to disable mime check on attachement upload

Hugome:
This add the ability to disable mime check on file upload.
(In my case I want to upload eml mail file and json metadata)

Changes in this pull request:

- Add a new env variable `DISABLE_UPLOAD_MIME_CHECK` to control the new behavior
- Edited `AttachmentHelper.php` to create that new behavior

@JC5

### DIFF
--- a/config/firefly.php
+++ b/config/firefly.php
@@ -283,6 +283,12 @@ return [
         'application/vnd.oasis.opendocument.formula',
         'application/vnd.oasis.opendocument.database',
         'application/vnd.oasis.opendocument.image',
+
+        /* EML */
+        'message/rfc822',
+
+        /* JSON */
+        'application/json',
     ],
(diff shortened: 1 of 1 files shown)


JC5:
It feels easier to extend the list with allowed mime types than add another option.

Hugome:
I was hesitating on that, but how could we do that ?
Creating a env variable like `ALLOWED_MIMES=application/json,message/rfc822,image/jpg` and split by `,` ?

@JC5

JC5:
No I mean, why does it have to be user configurable?

[This list](https://github.com/firefly-iii/firefly-iii/blob/main/config/firefly.php#L226) is pretty extensive. People rarely need new mime types. 

You could extend the list instead?

Hugome:
Oh I understand now, that's true that list can just be bigger if someone need others types.

I replaced my commit

@JC5

Issue #6122, opened 2022-06-01
type error on data import and display


nineff:
### Support guidelines

- [X] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)

### I've found a bug and checked that ...

- [X] ... [the documentation](https://docs.firefly-iii.org/) does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem
- [X] ... it's [definitely me, not you](https://github.com/firefly-iii/firefly-iii/blob/main/.github/its_you_not_me.md)

### Description

Importing a csv with the data importer results in the following message after starting the job to actually update the firefly-iii database:

The import routine has finished 🎉. Done!
Error(s)

`Line #0: Submission HTTP error: Status code is 500: {"message":"Internal Firefly III Exception: FireflyIII\\Support\\Steam::positive(): Argument #1 ($amount) must be of type string, int given, called in \/home\/niko\/firefly\/firefly-iii\/app\/Transformers\/TransactionGroupTransformer.php on line 120","exception":"TypeError"} `

The transactions are actually displayed correctly on the dashboard of firefly, but viewing details leads to further errors.

### Expected behaviour

- import should work
- in case of errors the "import routine has finished" message should be changed to " ... has failed" 
- imported data should be viewable

### Debug information

**❗I think I'm actually on FIDI version 0.9.10, but this is what the /debug page reports:**

Debug information generated at 2022-06-01 12:04:34 Europe/Berlin for Firefly III Data Importer version **0.9.9**.

| Scope | Version |
| --- | --- |
| FIDI | 0.9.9 |
| Build | (unknown), (unknown) |
| PHP | 8.1.6 |
| Host | Linux |

| System info | Value |
| --- | --- |
| Using docker? | false |
| System TZ | Europe/Berlin |
| Browser TZ | Europe/Berlin |
| App environment | local |
| App debug mode | false |
| App cache driver | file |
| App logging | debug, stack |
| Display errors | Off |
| Error reporting | ALL errors |
| Interface | cli-server |
| BCscale | 12 |
| Trusted proxies (.env) |  |

| User info | Value |
| --- | --- |
| User agent | Mozilla/5.0 (X11; Linux x86_64; rv:101.0) Gecko/20100101 Firefox/101.0 |

Debug information generated at 2022-06-01 12:29:31 Europe/Berlin for Firefly III version **5.7.9**.

| Scope | Version |
| --- | --- |
[...]

PR #6124, opened 2022-06-01
fiy type errors

nineff:
fix type errors

Fixes issue #6122

Changes in this pull request:

- added string type cast
- 
- 

@JC5


JC5:
Thanks for the effort! I've decided to fix this at the root, in the GroupCollector. Otherwise, there will be a billion different places to fix this. See my commit here: https://github.com/firefly-iii/firefly-iii/commit/f83ab0a7dc08845230e563e53a30517d87936b06

PR #6267, opened 2022-07-26
Add username to fit redis 6+ ACL capabilities

canoine:
This just adds the ability to set a redis username.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -114,7 +114,7 @@ REDIS_PORT=6379
 
 # Use only with Redis 6+ with proper ACL set. Leave empty otherwise.
 REDIS_USERNAME=
-REDIS_PASSWORD=null
+REDIS_PASSWORD=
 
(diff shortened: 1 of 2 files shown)


JC5:
Looks good. However, when it's empty or NULL, the config file will fallback to "default", even though your instructions say:

> Use only with Redis 6+ with proper ACL set. Leave empty otherwise.

Shouldn't that be `env('REDIS_USERNAME')`?

canoine:
You're right. Even more, perhaps forcing a default username will break the compatibility with Redis before 6.0.0.
I'll change the PR (sorry for the noise, I'm a beginner).

JC5:
Nice work, don't worry about the noise. As a final touch, could you change the password field in the .env.example file to `REDIS_PASSWORD=`? That way, it's nice and consistent.

canoine:
Done.

PR #6494, opened 2022-10-03
Develop

robvankeilegom:
Hey @JC5,

I made a change to the `firefly-iii:create-first-user` artisan command to also generate a personal access token. I'd like this change to be implemented so i can more easily run some integration tests for the paypal-importer repo is maintain.  I also commented out some route which linked to controllers that don't exist in the code base. 
- FireflyIII\Api\V1\Controllers\Autocomplete\RecurrenceController
- FireflyIII\Http\Controllers\Webhooks\IndexController
- FireflyIII\Http\Controllers\Json\ExchangeController

If you want me to change anything, lmk.

Thanks!

### DIFF
--- a/app/Console/Commands/CreateFirstUser.php
+++ b/app/Console/Commands/CreateFirstUser.php
@@ -81,7 +81,15 @@ class CreateFirstUser extends Command
         $user->save();
         $user->setRememberToken(Str::random(60));
 
+        // Make sure a passport client is available
+        $this->call('passport:client', ['--personal' => true, '--name' => config('app.name').' Personal Access Client']);
+        // Create a token for the user
+        $result = $user->createToken(
+            'testing', []
+        );
+
         $this->info(sprintf('Created new admin user (ID #%d) with email address "%s" and password "%s".', $user->id, $user->email, $password));
+        $this->info('Created the following access token: ' . $result->accessToken);
         $this->error('Change this password.');
 
         return 0;
--- a/routes/api.php
+++ b/routes/api.php
@@ -343,7 +343,7 @@ Route::group(
 
         Route::get('{recurrence}/transactions', ['uses' => 'ListController@transactions', 'as' => 'transactions']);
 // See reference nr. 16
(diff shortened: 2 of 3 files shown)


JC5:
Mmmm. I don't think I can merge this. Since this is also an artisan command, why not just call that command together with the "create-first-user" command?

(I appreciate the routes being commented away, but it should be in a separate PR, that makes it easier to manage. For now, it doesn't matter.)

robvankeilegom:
Ill split the PR up before merging and get the changes for the routes out. 

The artisan command `passport:client` isn't really necessary when working in a docker container since `passport:install` is called in the entrypoint script, which will install 2 passport clients. But i'm calling it for all other cases when a client might not be available. 

The part i need is the call to `createToken` since this is pretty much the same as requesting a token in Firefly under `Options > Profile > Oauth > Personal Access Tokens`.

If you don't like the change i'm assuming i can also get a token via the `oauth` routes added by Laravel Passport as an alternative. But since there's nothing documented about these routes in the api documentation and there's not way to to get the password client's secret and id via the interface, is assumed it wasn't officially supported. 

Thanks for your time!

JC5:
> If you don't like the change i'm assuming i can also get a token via the oauth routes added by Laravel Passport as an alternative. 

Yes, you can. Also as an alternative, you can also create a token and put the database records in your test database. That way, you always end up with the same test token.

Gaoura:
Hello guys ! Personal opinion on commenting the unused routes : delete them instead of commenting them out

Either these pieces of code are useful and you keep them, or they are useless and you delete them. If you ever need to find it again for some reason, Git let you search through the history of changes so I don't see any reason to keep some commented code

robvankeilegom:
I ended up inserting a known oauth client in the database and updating the random password from the `php artisan firefly-iii:create-first-user` command to something known. 

Then requesting a token with curl. 

`.drone.yml` file for anyone wondering: https://github.com/robvankeilegom/firefly-III-paypal-importer/blob/c2d9237d8710c539f6c910b68cea3b3db18e3f5e/.drone.yml

Thanks for the help!

JC5:
Ah that also works, thanks for sharing the details!

PR #6701, opened 2022-12-05
Ensure `remote_guard_alt_email` if changed

nebulade:
If the header value for email in proxy auth changes, also update in the database.

### DIFF
--- a/app/Support/Authentication/RemoteUserGuard.php
+++ b/app/Support/Authentication/RemoteUserGuard.php
@@ -84,7 +84,7 @@ class RemoteUserGuard implements Guard
             $emailAddress = (string) (request()->server($header) ?? null);
             $preference   = app('preferences')->getForUser($retrievedUser, 'remote_guard_alt_email');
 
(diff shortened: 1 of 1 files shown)


JC5:
Validating your code I found another issue. You could update your PR with the following, or I'll close it and submit it myself. Let me know! (you'll get the credits regardless)

```php
            if ('' !== $emailAddress && null === $preference && $emailAddress !== $userID) {
                app('preferences')->setForUser($retrievedUser, 'remote_guard_alt_email', $emailAddress);
            }
            // if the pref isn't null and the object returned isn't null, update the email address.
            if ('' !== $emailAddress && null !== $preference && $emailAddress !== $preference->data) {
                app('preferences')->setForUser($retrievedUser, 'remote_guard_alt_email', $emailAddress);
            }
```

- `$emailAddress` can't be null, that's a bug on my side.
- `$preference` is an object.

nebulade:
Thanks for looking at it, I guess it is easier if we close this here and you just put the code yourself, to avoid much back and forth. As mentioned I am no PHP developer :) I can test once the commits are pushed though.

Issue #5403, opened 2021-12-16
[data-importer] ING Germany description quirk


p-rintz:
### Support guidelines

- [X] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)

### I've found a bug and checked that ...

- [X] ... the documentation does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem

### Description

Im using Nordigen as the import for an ING Germany Account.

Sadly the transactions look like this:

`[2021-12-16 16:39:20] local.DEBUG: Submitting to Firefly III: {"apply_rules":true,"error_if_duplicate_hash":true,"transactions":[{"type":"withdrawal","date":"2021-11-30","amount":"999.0000000000000000","description":"mandatereference:,creditorid:,remittanceinformation:Wohngeld - Philipp Rintz","order":0,"currency_code":"EUR","tags":[],"category_name":null,"category_id":null,"external_id":"000007120123123123","internal_reference":"","source_id":11,"destination_name":"Removed for Ticket","destination_iban":"DE123123123123123123123123"}]}`

Which means that "mandatereference:,creditorid:,remittanceinformation:" as a String is prepended to every single transaction that gets imported.

Im not really sure if you can classify this as a bug, as its not really any bug from your side, but its hugely annoying when this gets imported. Feel free to close this if there is nothing you can/want to do.

I thought of creating a rule to remove this string from any transaction that appears, but this doesnt seem to be possible currently. So I guess unless you have another Idea, this might also be a feature request for creating that kind of rule.

### Expected behaviour

That *mandatereference:,creditorid:,remittanceinformation:* isnt being prepended to every transaction imported... ☹️ 

### Debug information

Debug information generated at 2021-12-16 16:46:30 Europe/Berlin for Firefly III Data Importer version **0.3.0**.

| Scope | Version |
| --- | --- |
| FIDI | 0.3.0 |
| PHP | 8.0.12 |
| Host | Linux |

| System info | Value |
| --- | --- |
| Using docker? | false |
| System TZ | Europe/Berlin |
| Browser TZ | Europe/Berlin |
| App environment | local |
| App debug mode | true |
| App cache driver | file |
| App logging | debug, stack |
| Display errors | Off |
| Error reporting | ALL errors |
| Interface | apache2handler |
| BCscale | 12 |
| Trusted proxies (.env) |  |

| User info | Value |
| --- | --- |
[...]

PR #6811, opened 2023-01-07
Implement new actions to remove strings from transaction's descriptions

Blogshot:
Changes in this pull request:

I have added three actions that can be used in automation -> rules to modify a transaction's description:
- Remove the first occurrence of a string
- Remove all occurrences of a string
- Remove using a RegEx-pattern

I've tested these actions locally and they seems to work robustly.

The trigger for these was https://github.com/firefly-iii/firefly-iii/issues/5403, where ING Germany in combination with Nordigen contains `mandatereference:,creditorid:,remittanceinformation:` in front of the description.

I'm not a professional programmer, so please let me know if there is something that needs to be added to the PR.

### DIFF
--- a/app/TransactionRules/Actions/RemoveFirstFromDescription.php
+++ b/app/TransactionRules/Actions/RemoveFirstFromDescription.php
@@ -1,0 +1,64 @@
+<?php
+/**
+ * RemoveFirstFromDescription.php
+ * Copyright (c) 2023 sascha@knott.ac
+ *
+ * This file is part of Firefly III (https://github.com/firefly-iii).
+ *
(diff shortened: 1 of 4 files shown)


Blogshot:
Maybe relevant: I based these files on `app/TransactionRules/Actions/PrependDescription.php` because it also modifies descriptions.

I haven't seen any, but it may be that there is some unique/unnecessary code that I've copied to my actions.

JC5:
Although I appreciate what you built, I'm not going to merge this. In line with my goals and ideas for the rule engine, I don't want to add more complexity than it has already. 

https://docs.firefly-iii.org/firefly-iii/about-firefly-iii/what-its-not/

PR #6826, opened 2023-01-10
Hide sensitive information in 2FA setup page when printing

jstebenne:
The QR code (and manual code) should not be recoverable after the initial setup. This would allow an unauthorized person to access an account without leaving a trace (like showing that a backup code was used, given that person has the account password).

Even if very low, having that information visible could be a problem.

---
Changes in this pull request:

- Add CSS classes to hide sensitive information when printing the 2FA setup page
- Hide instructions on how to use the QR code
- Remove dead code

Previews:

@JC5


jstebenne:
I could also hide the `Code` field at this point. I wasn't sure if it would be necessary.

JC5:
Nice, merged.

PR #6942, opened 2023-01-29
Correct Norwegian translation for auto budgetting

aleksen:
Fixes issue # (if relevant)

Changes in this pull request:

- Changed translation for one of the auto budgetting types in Norwegian. The biggest problem was that the two different types had the same translation. Now they don't. New translation uses the term "Legg til", which means "Add". Original translation was "Angi" which means "Set"
- 
- 

@JC5


JC5:
Thanks for this!

Please submit translations to: https://crowdin.com/project/firefly-iii

Otherwise a new release will simply overwrite them.

PR #7001, opened 2023-02-08
Add week days to Rule Triggers

miguelangel-nubla:
Signed-off-by: miguelangel-nubla <miguelangel.nubla@gmail.com>

### DIFF
--- a/app/Support/ParseDateString.php
+++ b/app/Support/ParseDateString.php
@@ -25,6 +25,7 @@ declare(strict_types=1);
 namespace FireflyIII\Support;
 
 use Carbon\Carbon;
+use Carbon\CarbonInterface;
 use FireflyIII\Exceptions\FireflyException;
 use Log;
 
@@ -46,6 +47,13 @@ class ParseDateString
             'end of this quarter',
             'start of this year',
             'end of this year',
+            'monday',
+            'tuesday',
+            'wednesday',
+            'thursday',
+            'friday',
+            'saturday',
+            'sunday',
         ];
 
     /**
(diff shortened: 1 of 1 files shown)


JC5:
What do you expect these triggers to do?

miguelangel-nubla:
[Transaction|Payment day|Due date|other date] is on [weekday]
Screenshot 2023-02-09 153045

JC5:
The problem is that this will trigger when *today* is Monday, not when the transaction is on Monday. This is less of a problem for the other custom searches.

miguelangel-nubla:
I see your point.

It currently triggers when _today's week [weekday]_ is equal to the date.
So "monday of this week", "tuesday of this week" trigger values would make more sense?

JC5:
>  So "monday of this week", "tuesday of this week" trigger values would make more sense?

The current set of text triggers ("start of this quarter" etc) do not make sense in the context of rules, because they apply to today's date, not to the transaction's date. So expanding this doesn't really work right now.

It will not be useful in a rule. "Monday of this week" seems a strange filter when you're importing months of transactions. All those months will not trigger the rule, you can't set "any Monday", it's just "Monday of this week".

miguelangel-nubla:
I agree any "monday" would be a more useful filter.

From what I can tell with a quick look at the code, all the date operators for the search would need to be modified, as they compare against exact dates, or date rages, none of which apply:
```
date_is:2020-01-01
date_before:today
date_after:"start of this month"
interest_date_* (same options)
book_date_* (same options)
process_date_* (same options)
due_date_* (same options)
payment_date_* (same options)
invoice_date_* (same options)
created_at_* (same options)
updated_at_* (same options)
```

Do you have any pointers where would be a good place to implement this?

JC5:
> Do you have any pointers where would be a good place to implement this?

Some time in the future. This needs a whole bunch of changes, and will either break the search engine, or the rule engine. I'd like to postpone that because other things are more important right now.

PR #7535, opened 2023-05-22
Withdraw from piggy banks

ChrisWin22:
Fixes issue # 6631

Changes in this pull request:

- Added ability to deduct from a Piggy Bank on a Withdrawal
- Added ability to add to Piggy Bank from a Deposit

@JC5

The feature to add to a Piggy Bank from a Deposit was not a part of the original issue but when implementing it seemed to make sense to do it at the same time. If you would like me to make it only for Withdrawals let me know and I can change that.

### DIFF
--- a/app/Console/Commands/Correction/FixPiggies.php
+++ b/app/Console/Commands/Correction/FixPiggies.php
@@ -25,11 +25,10 @@ namespace FireflyIII\Console\Commands\Correction;
 
 use FireflyIII\Models\PiggyBankEvent;
 use FireflyIII\Models\TransactionJournal;
-use FireflyIII\Models\TransactionType;
 use Illuminate\Console\Command;
 
 /**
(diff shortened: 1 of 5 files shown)


JC5:
Looks good. Could you drop the generated JS code from the PR though?

ChrisWin22:
Of course! I wasn't sure if you wanted them or not.

JC5:
Also check out:

https://github.com/firefly-iii/firefly-iii/blob/develop/app/TransactionRules/Actions/UpdatePiggybank.php#L68

JC5:
Nice. I might add a few details to the FixPiggies command, to make sure piggy bank events are only linked to transfers, withdrawals or deposits. But i'm not sure how that works in your PR so i'll commit that separately. It has no priority.

Good work! Merged :+1:

PR #7676, opened 2023-06-20
Rule ConvertToDeposit: fix missing parameter

timendum:
Fixes issue # (if relevant)

Changes in this pull request:

- a minor fix on the ConvertToDeposit, a missing parameter in the where condition.

When the rules triggers, I'm getting this error:

    Internal Firefly III Exception:
    SQLSTATE[22007]:
    Invalid datetime format: 1292 Truncated incorrect DOUBLE value: '=' (Connection: mysql,
    SQL: update `transactions` set `account_id` = 254 where `transaction_journal_id` = = and `amount` < 0
    )

I hope this fix it, I've also checked others ConvertToX rules, but they look good to me.

@JC5


JC5:
Nice find, that should fix it! 👍🏻

PR #7717, opened 2023-07-04
Cover the startOfPeriod method of Navigation class

tonicospinelli:
Just continue the work to increase the coverage of Navigation class

@JC5


JC5:
Nice :+1:

PR #8028, opened 2023-10-05
Fix interest payment deducted from amount due

tieu1991:
Changes in this pull request:

Withdrawal from liability to expense account currently fall in all other case, which decrease the debt. This can be interest payment, so I added a case to increase de debt for this type of transaction.

@JC5


JC5 (approved):

JC5:
:+1:

Issue #8069, opened 2023-10-17
Newly created rules have "1" as description


avee87:
### Support guidelines

- [X] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)

### I've found a bug and checked that ...

- [X] ... [the documentation](https://docs.firefly-iii.org/) does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem
- [X] ... it's [definitely a Firefly III issue, not me](https://github.com/firefly-iii/firefly-iii/blob/main/.github/its_you_not_me.md)

### Description

This is a minor but annoying issue. For some reason when you create a rule with empty description it ends up with "1" which needs to be then manually removed.
Tested and reproduced on demo instance.
<img width="1116" alt="Screenshot 2023-10-17 at 17 48 32" src="https://github.com/firefly-iii/firefly-iii/assets/6134677/4dc385a1-77e0-406f-8699-4ad396d84bc9">

### Debug information

Debug information generated at 2023-10-17 17:44:43 for Firefly III version **v6.0.27**.

[...]

PR #8070, opened 2023-10-17
[Issue #8069] Fixed bug on description field

Maxco10:
Fixes issue https://github.com/firefly-iii/firefly-iii/issues/8069

Changes in this pull request:
- During the refactoring the description field was lost. Now, I update the RuleRepositories->store function with right code for description field.

@JC5

### DIFF
--- a/app/Repositories/Rule/RuleRepository.php
+++ b/app/Repositories/Rule/RuleRepository.php
@@ -289,7 +289,7 @@ class RuleRepository implements RuleRepositoryInterface
         $rule->strict          = array_key_exists('strict', $data) ? $data['strict'] : false;
         $rule->stop_processing = array_key_exists('stop_processing', $data) ? $data['stop_processing'] : false;
         $rule->title           = $data['title'];
-        $rule->description     = array_key_exists('stop_processing', $data) ? $data['stop_processing'] : null;
+        $rule->description     = $data['description'];
         $rule->save();
         $rule->refresh();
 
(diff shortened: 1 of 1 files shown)


JC5:
A rule does not need to have a description, could you make sure a similar check to the title is added?

Maxco10:
@JC5 sorry, but i did not get your comment. Please read the issue.

app/Repositories/Rule/RuleRepository.php
```diff
@@ -289,7 +289,7 @@ public function store(array $data): Rule
         $rule->strict          = array_key_exists('strict', $data) ? $data['strict'] : false;
         $rule->stop_processing = array_key_exists('stop_processing', $data) ? $data['stop_processing'] : false;
         $rule->title           = $data['title'];
-        $rule->description     = array_key_exists('stop_processing', $data) ? $data['stop_processing'] : null;
+        $rule->description     = $data['description'];
```
JC5, line 292: This field can be empty or non-existent. So I would expect this instead:

```
 array_key_exists('description', $data) ? $data['description'] : ''';
```

JC5 (changes requested):

Maxco10:
@JC5 PR Updated. I added check on title field, if the title does not exists, we set a default value "untitle", similar case for description field but the default value for this case is "" an empty string.

JC5:
Could you fix the spelling of "Untitled" and use single quotes instead of double?

Maxco10:
@JC5 Done!

PR #8081, opened 2023-10-20
Cover endOfPeriod method of the Navigation class

tonicospinelli:
The custom frequency requires a timezone configuration, forcing it to run in the integration test scope.

Running the integration tests requires a database connection in the build process. It enables other case tests.

The API Tests cause interference in other tests, requiring isolating them.

Fixes issue # (if relevant)

Changes in this pull request:

-
-
-

@JC5

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -1,14 +1,37 @@
 name: Sonarcloud
 on:
-  pull_request: 
+  pull_request:
   push:
     branches:
       - main
       - develop
+env:
+  DB_CONNECTION: mysql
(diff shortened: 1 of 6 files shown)


JC5:
Nice work, let me know when it's ready to merge!

JC5:
Is it an idea to use SQLite instead of MySQL?

tonicospinelli:
Interesting question! It is better to use tools like in production for the integration tests to assert the application is ready.

The next step is to build the integration tests using any kind of database available to be integrated with the application.

For example, building the application and running the integration test over MySQL, Postgres, SQLite, and other options will be nice.

PR #8084, opened 2023-10-22
Hide duplicated webhook create button

JoSchrader:
On the webhooks index page there are two "create new webhook" buttons. One above and one below the table listing the webhooks. When there are no webhooks both buttons are right next to each other which looks kind of like a mistake.

Changes in this pull request:
- Hide duplicated webhook create button when there are no webhooks

Empty table before:
before-empty
Empty table after:
after-empty

<details>
  <summary>With existing webhooks (before and after):</summary>
  
</details>

@JC5

### DIFF
--- a/resources/assets/js/components/webhooks/Index.vue
+++ b/resources/assets/js/components/webhooks/Index.vue
@@ -80,7 +80,7 @@
             </tbody>
           </table>
 
-          <div style="padding:8px;">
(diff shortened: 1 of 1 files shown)


JC5:
Could you drop the compiled JS from the PR?

JoSchrader:
Sure, dropped the commit that added the compiled js

JC5:
Nice work, that looks better indeed.

PR #8123, opened 2023-11-03
increase update check timeout from ~3s to ~10s

RoguedBear:
## Context
I self host firefly on my own hardware, and for a long while I've been seeing the "connection timed out after 3141 milliseconds" when firefly checks for an update. I thought the internet in the container was broke however after curling the update endpoint inside the container, i realised that the shorter curl timeout was causing the update check to timeout.

Therefore i have increased the timeout from short 3 seconds to a moderate 10s.

(i did observe the original timeout was pi, so I increased it by a multiple of 3 :P )

## Changes in this pull request:

- increases update check timeout from 3s (π) to 10s (3*π)

@JC5

### DIFF
--- a/app/Services/FireflyIIIOrg/Update/UpdateRequest.php
+++ b/app/Services/FireflyIIIOrg/Update/UpdateRequest.php
@@ -86,7 +86,7 @@ class UpdateRequest implements UpdateRequestInterface
                 'headers' => [
                     'User-Agent' => sprintf('FireflyIII/%s/%s', config('firefly.version'), $channel),
                 ],
-                'timeout' => 3.1415,
+                'timeout' => 9.4245,
             ];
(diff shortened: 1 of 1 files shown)


RoguedBear:
I am not sure why it takes more than 3s. 

I did some testing, inside the container it took 4s-6s to get the versions data, but on the host it was instantaneous. Running the same command on a different container, the result was instantaneous too.

But i do not think the issue is in the image, as i tried this experiment on a cloud vm too and the timing was instantaneous too inside the container.

Given these variations i thought raising the timeout to not something too big, but not something short either could cover unlikely edge cases

JC5:
I already increased it to 3 seconds to catch most issues. Any higher, and it takes too long to realize it's timing out. I'm afraid this one is for you to fix, not me.

PR #8219, opened 2023-12-01
Fixed downlodWebhok function after adding a new webhook.

Maxco10:
Fixes issue:
After adding a new webhook the showed list is empty because the relative path is wrong in case of subfolder on main domain.

Ex:
the domain where fireflyIII is my-site.net/public
After adding a new webook the list will retrieve from my-site.net and not from my-site.net/public

Changes in this pull request:

- Added "." as currenct directory othrwise we get a bad relative path.

@JC5

### DIFF
--- a/resources/assets/js/components/webhooks/Index.vue
+++ b/resources/assets/js/components/webhooks/Index.vue
@@ -122,7 +122,7 @@ export default {
       webhook.show_secret = !webhook.show_secret;
     },
     downloadWebhooks: function (page) {
-      axios.get("/api/v1/webhooks?page=" + page).then((response) => {
+      axios.get("./api/v1/webhooks?page=" + page).then((response) => {
         for (let i in response.data.data) {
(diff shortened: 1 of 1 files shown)


JC5:
I want to suggest you fix your /public path instead.

Maxco10:
Yep, good but I used "public" as example, the issue is related to any subfolder used.
This piece of code on 6.0.30 is working if only if the firefly root is on main root of server.

JC5:
Sure you did lol.

Maxco10:
I added some censured attachements. As you can see, I do not have "public" folder opened but I have fireflyIII project in subfolder. The redirect after adding a new webook is wrong because it does not follow the correct path without "."

Please, take some minutes to double check this my pull request or this issue.

Screenshot 2023-12-06 at 12 39 22
Screenshot 2023-12-06 at 12 30 54

@JC5

Issue #6432, opened 2022-09-12
Rules > new "Description matches regex" trigger


tomaskir:
### Support guidelines

- [X] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)
- [X] My request is not listed as [a very good idea, but unfortunately...](https://docs.firefly-iii.org/firefly-iii/about-firefly-iii/what-its-not/)
- [X] I've used [the search](https://github.com/firefly-iii/firefly-iii/issues?q=is%3Aissue) and this has not been discussed before.

### Description

In Rules, the current `Description ...` triggers are not flexible enough, and require many duplicate rules to be created in case the description is the only difference.

A single new trigger `Description matches regex` would help a lot to cut down on duplicate rules.

### Solution

Add a new trigger `Description matches regex`

### What are alternatives?

This has been discussed in #1281 as a general regex support, but this request is not that complex. The Strict vs. Non-strict mode (#1338) is not a good solution, since it's global to all rules.

### Additional context

This small trigger addition would allow me to cut down the number of rules I have by about a half. Since it's just a single new trigger among many others already existing, I don't think it violates any principles of Firefly.

JC5:
https://docs.firefly-iii.org/firefly-iii/about-firefly-iii/what-its-not/#more-complicated-rules-like-and-or-or-copypasting-values-etc

tomaskir:
> https://docs.firefly-iii.org/firefly-iii/about-firefly-iii/what-its-not/#more-complicated-rules-like-and-or-or-copypasting-values-etc

With all due respect - I don't think this applies to a request for a new trigger. This is not a request to add complicated logic or to change how Rules work in Firefly. This is a request for a single new trigger.

> For Firefly III, the current features of the engine will not be expanded. If you have a suggestion, please consider that your suggested trigger or action is simple.

Considering there are already triggers like `Description starts with` and `Description ends with`, I think a `Description matches regex` trigger is within the Firefly philosophy on how Rules work.

JC5:
Sorry, it's not going to happen.

ohare93:
> The rule engine could be a lot more complex. Things like conditional rules, if-then-else, regular expressions or variables. These are all technically possible. **However, the rule engine must also be usable and easy to be introduced to.** Most rule engines, even the filters in GMail, are very complex because of user demands and as a result, nobody uses them any more.

This explanation is incredibly strange and unsatisfying while also suggesting users are dumb.

[...]

PR #8220, opened 2023-12-01
Adding regexp to description trigger

enboig:
I wanted a trigger which assign a category if:
- movement category is empty
- description contains "lorem" or "ipsum

I know it is an edge case, so in order to not complicate things for other uses, adding regexp allow a solution.

I read about including translations after committing it, I hope it isn't a problem.

First merge request I have made, sorry if I missed something. And now I am a symfony developer, I hope my code is ok.

### DIFF
--- a/app/Helpers/Collector/GroupCollector.php
+++ b/app/Helpers/Collector/GroupCollector.php
@@ -1018,6 +1018,42 @@ class GroupCollector implements GroupCollectorInterface
     }
 
     /**
+     * Use regexs to filter descriptions.
+     *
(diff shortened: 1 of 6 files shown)


enboig:
After reading https://github.com/firefly-iii/firefly-iii/issues/6432#issuecomment-1279375024 I thought there was no problem as long as somebody else implemented it.

enboig:
I understand you wanting to avoid complexity in filters, but in my case what could be solved on a single rule, has become 9 rules.

So I am asking you to allow me improve your app.

PR #8348, opened 2024-01-04
Update outdated documentation URLs in .env.example

lemuelroberto:
Changes in this pull request:

- Update outdated documentation URLs in .env.example

image

@JC5


JC5:
Nice, I completely forgot about those.

PR #8432, opened 2024-01-21
Update favicons.twig

stevietv:
Changes in this pull request:

When I use firefly-iii behind oauth2-proxy which handles authentication, it sets a cookie.
With this cookie, requests are forwarded to the firefly-ii.
Everything works fine, except for the manifest.json, where the browser doesn't send the cookie and you end up with an error (CORS).

The solution is to add the crossorigin="use-credentials" attribute to the tag:

<link rel="manifest" href="/manifest.json" crossorigin="use-credentials">

Documentation :
[crossorigin](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin)
[link attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-crossorigin)

@JC5


JC5:
Nice, merged!

PR #8650, opened 2024-03-08
[feat] Rules Expression Engine

michaelhthomas:
# Changes in this PR

Adds an expression engine for rule action values based on `symfony/expression-language`. Any action value prefixed with an `=` sign will be treated as a calculated expression, which will be evaluated for each transaction and used as the action value for that transaction. Expressions are validated on rule update, ensuring that rules with invalid expressions cannot be saved.

# Motivation

This change is beneficial for a large variety of powerful use cases, or even some fairly simple ones. A fairly simple example can be seen with processing imported Venmo transactions, and automatically setting the source account for deposits:

- A Venmo deposit transaction will have a description of the form: "Jane Doe Paid John Smith"
- Imported Venmo transactions can be matched by the following triggers:
	- Destination account name is: Venmo
	- Description ends with: Paid John Smith
- Using an action with a simple expression, the source account can be set automatically:
	- Set source account to: `=substr(description, 0, strlen(description) - strlen(" Paid John Smith"))`

An even simpler example would be as a replacement for the "Prepend description with" action. Instead, we may simply use the "Set description to" action with an expression:

- Original Action
	- Prepend description with: `foo`
- New Action
	- Set description to: `="foo"~description`

This change would thus reduce the number of separate actions that have to be developed in the future to support more advanced behavior, instead enabling users to create complex actions while reducing maintenance burden.

# Screenshots

## Successful Automation

image

## Validation Message

image

# Disclaimer

I am aware that [there is some reluctance](https://docs.firefly-iii.org/explanation/more-information/what-its-not/#more-complicated-rules-like-boolean-logic-copypasting-values-etc) to increasing the complexity of the rules engine. However, I would like to make the case that adding support for expressions in this way actually offers the potential to decrease the complexity of the rules engine moving forward, and the implementation is extremely simple. 

All the expression logic in this PR is handled by `synfony/expression-language`, a well-maintained and documented component of the Symfony framework that is easily extensible, but comes with sane defaults and a familiar expression syntax. The implementation in this PR is extremely simple, and should be straightforward to maintain and support going forward.

In fact, this change could enable the removal of some rules in the future, such as the "Append" and "Prepend" actions, which could be easily replaced by simple expressions as demonstrated above, reducing the currently (slightly) overwhelming nature of the actions list.

I am also willing to take on a share of this burden and engage in writing support documentation for this feature, and making further changes to better integrate with Firefly III. I believe that the ability for users to implement more complex rules is extremely important, and that this change comes without increased complexity for beginner users, and little new code in need of maintenance.

@JC5


Fmstrat:
Wouldn't regex matches be more universal? Don't mean to downplay the PR, just knowing regexes with group 1 would be pretty straight forward, too. I.e: `/^([a-z0-9\/\.\-]) Paid John Smith/i`

JC5:
Regex will not be happening.

michaelhthomas:
I'd argue regex matching with groups passed as arguments to the actions would actually be more complicated to grasp while offering less functionality.

Since all transaction data is exposed to expressions, you can use data in actions without having to match it in triggers. The simple appending to the description example, for one, would not be possible/straightforward with just regex matching.

JC5:
I'm going to spend some time on this on the weekend, it looks pretty cool.

app/Api/V1/Requests/Models/Rule/ValidateExpressionRequest.php
```diff
@@ -0,0 +1,46 @@
...
+namespace FireflyIII\Api\V1\Requests\Models\Rule;
+
+use FireflyIII\Support\Request\ChecksLogin;
+use FireflyIII\Support\Request\ConvertsDataTypes;
+use Illuminate\Foundation\Http\FormRequest;
+
+/**
+ * Class ValidateExpressionRequest
+ */
+class ValidateExpressionRequest extends FormRequest
```
JC5, line 34: Add a `rules()` function that returns something like `['expression' => 'required|ruleActionExpression']`

app/Api/V1/Requests/Models/Rule/StoreRequest.php
```diff
@@ -123,7 +123,7 @@ public function rules(): array
             'triggers.*.stop_processing' => [new IsBoolean()],
             'triggers.*.active'          => [new IsBoolean()],
             'actions.*.type'             => 'required|in:'.implode(',', $validActions),
-            'actions.*.value'            => 'required_if:actions.*.type,'.$contextActions.'|ruleActionValue',
```
JC5, line 126: Instead of adding a method to FireflyValidator, create a new class (modeled after `IsBoolean` for example) in `app/Rules`, and the convert the rule to something like:

```php
'actions.*.value' => ['required_if:actions.*.type,'.$contextActions, new IsRuleActionExpression(), 'ruleActionValue']
```

app/Api/V1/Requests/Models/Rule/ValidateExpressionRequest.php
JC5: See also the comment about the class/method for validation.

app/Api/V1/Requests/Models/Rule/UpdateRequest.php
```diff
@@ -140,7 +140,7 @@ public function rules(): array
             'triggers.*.stop_processing' => [new IsBoolean()],
             'triggers.*.active'          => [new IsBoolean()],
             'actions.*.type'             => 'required|in:'.implode(',', $validActions),
-            'actions.*.value'            => 'required_if:actions.*.type,'.$contextActions.'|ruleActionValue',
+            'actions.*.value'            => 'required_if:actions.*.type,'.$contextActions.'|ruleActionExpression|ruleActionValue',
```
JC5, line 143: See comment in StoreRequest

app/Http/Requests/RuleFormRequest.php
```diff
@@ -147,7 +147,7 @@ public function rules(): array
             'triggers.*.type'  => 'required|in:'.implode(',', $validTriggers),
             'triggers.*.value' => sprintf('required_if:triggers.*.type,%s|max:1024|min:1|ruleTriggerValue', $contextTriggers),
             'actions.*.type'   => 'required|in:'.implode(',', $validActions),
-            'actions.*.value'  => sprintf('required_if:actions.*.type,%s|min:0|max:1024|ruleActionValue', $contextActions),
+            'actions.*.value'  => sprintf('required_if:actions.*.type,%s|min:0|max:1024|ruleActionExpression|ruleActionValue', $contextActions),
```
JC5, line 150: See comment in StoreRequest

app/Models/RuleAction.php
```diff
@@ -93,4 +94,10 @@ protected function ruleId(): Attribute
             get: static fn ($value) => (int)$value,
         );
     }
+
+    public function getValue(array $journal)
```
JC5, line 98: Will need a return value

JC5: What happens when the action value is NOT an expression? Not sure of the method name either, perhaps something like "getExpressionValue"?

app/TransactionRules/Expressions/ActionExpression.php
```diff
@@ -0,0 +1,150 @@
...
+declare(strict_types=1);
+
+namespace FireflyIII\TransactionRules\Expressions;
+
+use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
+use Symfony\Component\ExpressionLanguage\SyntaxError;
+
+class ActionExpression
+{
+    private static array $NAMES = array(
```
JC5, line 32: Fields that can be safely dropped are:

- deleted_at, encrypted

What's the deal with all the `*_count` fields?  Who populates them?

app/Validation/FireflyValidator.php
```diff
@@ -253,6 +254,27 @@ public function validateMustExist($attribute, $value, $parameters): bool
         return 1 === $count;
     }
 
+    public function validateRuleActionExpression(string $attribute, string $value = null): bool
```
JC5, line 257: See earlier comment about rule validators.

composer.lock
JC5, line 1: Please drop composer.lock changes.

routes/api.php
```diff
@@ -606,6 +606,8 @@ static function (): void {
         Route::put('{rule}', ['uses' => 'UpdateController@update', 'as' => 'update']);
         Route::delete('{rule}', ['uses' => 'DestroyController@destroy', 'as' => 'delete']);
 
+        Route::post('validateExpression', ['uses' => 'ExpressionController@validateExpression', 'as' => 'validate']);
```
JC5, line 609: `::post('validate-expression', ...`

resources/lang/en_US/validation.php
```diff
@@ -46,6 +46,7 @@
     'reconciled_forbidden_field'      => 'This transaction is already reconciled, you cannot change the ":field"',
     'deleted_user'                    => 'Due to security constraints, you cannot register using this email address.',
     'rule_trigger_value'              => 'This value is invalid for the selected trigger.',
+    'rule_action_expression'          => 'Invalid expression. :error',
```
JC5, line 49: where are the translation strings for the syntax errors?

JC5 (changes requested):

resources/lang/en_US/validation.php
michaelhthomas: These are unfortunately hard-coded in the expression parser's [`SyntaxError` class](https://github.com/symfony/expression-language/blob/7.0/SyntaxError.php) and within the lexer. I doubt there will be many changes to the format of these error messages, so it should be possible to parse the data out of them and then create translatable messages.

app/Models/RuleAction.php
michaelhthomas: The `ActionExpression` class just ignores any value that is not prefixed with an `=` sign, so the call to `evaluate` will just return the value as passed in. So `getValue` just returns the `action_value` if it's not an expression, or evaluates it and returns the result if it is an expression.

app/TransactionRules/Expressions/ActionExpression.php
michaelhthomas: I grabbed these from the `TransactionJournal` class, but turns out they were totally wrong. I've updated them to reflect the actual fields present on the `$journal` array.

app/Api/V1/Requests/Models/Rule/ValidateExpressionRequest.php
michaelhthomas: Took a stab at doing it this way, still not sure how I feel about it. Please let me know if there's a better way to do this.

composer.lock
michaelhthomas: There were some necessary changes since the `symfony/expression-language` dependency was added, but I've dropped all the random formatting changes it had strewn everywhere.

JC5 (approved):

JC5:
Nice work. I'm going to comb through this one last time and update the documentation.

michaelhthomas:
Thanks! I have two remaining possible concerns before this is released.

1. If anyone has an action value which begins with `=` (for some reason) those will break. If necessary, I should be able to write a migration to fix this.
2. Currently, errors in expressions (like accessing properties of variables which don't exist) will throw an uncaught exception, leading to a 500 and stopping further rule execution. Not sure what the expected behavior is here, but it might be better to just log the error to the console and audit log and skip past. 

I'm happy to make fixes for either of these if necessary. Thanks for your help getting this through, and please let me know if you need any help with the documentation!

JC5:
Nice. For the second one a fix is easy, I'll brood on the first one. 

It needs an escape character maybe, plus a migration. But a migration may be necessary for more parts of this feature, like the rule action which aren't necessary anymore.

JC5:
A migration (for both issues) is now in place.

https://github.com/firefly-iii/firefly-iii/commit/b3649cd4d07dbb612776a0d3d0d4c18d04eea2b8

PR #8719, opened 2024-03-26
Added unix_socket to pgsql

charmoniumQ:
Postgres serves over TCP sockets and UNIX sockets. Some users may prefer UNIX sockets because they can use their UNIX user as authentication to the database. Firefly-iii already supports MySQL over UNIX sockets and a very small change will allow Firefly-iii to support Postgres over UNIX sockets.

@JC5


JC5:
Given your comment [here](https://github.com/orgs/firefly-iii/discussions/8395#discussioncomment-8917494), is this a solution that works?

charmoniumQ:
No, this PR is in draft because I couldn't actually get it to work :/

I'll comment again if I can get it to work; for now I just switched to TCP and password authentication.

PR #8746, opened 2024-04-02
Set date to now when cloning journal

imlonghao:
Fixes issue # None

Changes in this pull request:

- Set date to now when cloning journal

Commit https://github.com/firefly-iii/firefly-iii/commit/c979cfcd89457188976b2a9df04b22c02fa26e88 change the date from `Carbon::now()` to `today(config('app.timezone'))` when cloning journal

I believe when we clone a transaction, we hope the date to be now, not the start of today.

@JC5


JC5:
Looks good, merged!

Issue #2781, opened 2019-10-29
[F-R] Add `DB_Socket` to overall config.  


Herve-M:
Hello,

**Description**
It would be cool if the `DB_Socket` setting is acessible and usable, in case of the setup with Docket but (metal bar) database, sharing a socket is the easiest and maybe secure as sharing all local port. 

**Solution**
Seem that mostly everything is settup [here](https://github.com/firefly-iii/firefly-iii/blob/f4b9f81e7149cdafa951b0c9251cdfb7506f2177/config/database.php#L58) but it miss the config. test as: if DB_Socket is set, not need of DB_Host + DB_Hostname.

**Additional context**
For Docker context, with a local DB, it make it easier to setup.
For normal setup, access through a socket could me more easy to secure on the system. 

Thanks,
Hervé-M

JC5:
As far as I can tell from Laravel's code, it's enough to set DB_SOCKET and leave DB_HOST empty. Did you try that already?

PR #8853, opened 2024-05-02
Add support for PostgreSQL unix socket connection

stackcoder:
This PR fixes issue #2781, #8395

Changes in this pull request:
- Add config option `DB_SOCKET` also to pgsql (likely mysql)

@JC5

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -98,6 +98,7 @@ return [
             'database'    => envNonEmpty('DB_DATABASE', $database),
             'username'    => envNonEmpty('DB_USERNAME', $username),
             'password'    => env('DB_PASSWORD', $password),
+            'unix_socket' => env('DB_SOCKET', ''),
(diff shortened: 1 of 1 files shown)


JC5:
So, does it work when you do this?

stackcoder:
Sorry for bothering you, there's no change needed!

I got it working by just setting `DB_HOST` to the `unix_socket_directory` of PostgreSQL as described in [php docs: pg_connect](https://www.php.net/manual/en/function.pg-connect.php#93837). The full path to the socket gets automatically expanded respecting the `DB_PORT` variable.

So a full configuration could look like that:

```nix
DB_CONNECTION = "pgsql";
DB_HOST = "/run/postgresql";
DB_PORT = 5432;
DB_DATABASE = "firefly-iii";
DB_USERNAME = "firefly-iii";
```

Issue #4043, opened 2020-11-08
What does "what's playing?" mean?


somasis:
This is a pointless question, but I was wondering, what does "what's playing" actually mean? It doesn't really make sense to me in the context of the dashboard, so I was wondering if there's some sort of inside joke or some other meaning to it.

JC5:
It’s just like “what’s going on?”

somasis:
Huh. Is that regional at all? I've never heard of that (coming from a US perspective) I guess.

PR #8906, opened 2024-05-21
Introduce new 'title' l18n key

chevdor:
This PR introduces a small fix that makes the title a title.

The current "What's playing" [does not mean much for most people]( https://github.com/firefly-iii/firefly-iii/issues/4043).
It may sound cool (assuming it is understood) in some context but it is confusing as title.

This PR leaves `welcome back` translated as `What's playing` but introduced a new key `title` that should remove some of the confusion.

Until I found the issue linked above, I thought this was a bug or a browser glitch. `What's playing ?` looks like the tab is about some media player... that would be playing something... Even translated as `Welcome back`, this would be rather useelss as title and anything like `FF III`, `FF`, `Firefly`, Firefly III` would be better.

I have zero opinion on what the default content should be, as long as it helps identity what this browser tab is about.

This PR fixes issue #4043.

Changes in this pull request:

- add a new key `title`
- switch subtitle from `welcome_back` to the new `title`
- set `title` to `Firefly III` byt default

@JC5


chevdor:
I did expect this kind of resistance so it is with little surprise.
The current title, nonetheless, makes no sense at all.

JC5:
That's fair. Feel free to open a discussion.

PR #8935, opened 2024-05-30
bcadd (string) fixes for SQLite float values

0xSylice:
As a brand new user to Firefly, after just setting up first install - using SQLite as database, creating my first user and submitting my first bank balance, I immediately encountered a crash with the:

`Argument #2 ($num2) must be of type string, float given` error

There is not a specific PR that I am aware of for this particular one but there are a few related issues. After looking at how they were fixed I see that the float values were just casted to strings to strings to make them compatible with bcadd().

See similar previous issues here: https://github.com/firefly-iii/firefly-iii/issues/6260 https://github.com/firefly-iii/firefly-iii/issues/8907

I implemented the same fix. Reinstalled my Firefly again from scratch, followed same process, created new user + new account with balance and the crash no longer occurs.

### DIFF
--- a/app/Support/Models/AccountBalanceCalculator.php
+++ b/app/Support/Models/AccountBalanceCalculator.php
@@ -120,13 +120,13 @@ class AccountBalanceCalculator
 
             // first create for normal currency:
             $entry               = $this->getAccountBalanceByAccount($account, $transactionCurrency);
-            $entry->balance      = bcadd((string) $entry->balance, $sumAmount);
+            $entry->balance      = bcadd((string) $entry->balance, (string) $sumAmount);
             $entry->save();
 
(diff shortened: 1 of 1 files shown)


JC5:
Hi there!

This issue is already fixed in the develop branch. Notice the cast to string a few lines above your fix (twice).

0xSylice:
Alright, I did not test the development branch install first before applying this, I just applied it to the current release that I was installing.

If it is not necessary to cast the other values for this fix to fully work, feel free to close this request. I simply applied the casting to all bcadd() parameters.

JC5:
Which is the correct solution! Thanks for your PR. I will close it, but not before expressing my appreciation 😊

PR #8951, opened 2024-06-04
add icon for delete action

stevewasiura:
in workflow of reconciling account, sometimes a record needs to be deleted (i.e. in event of a duplicated transaction from an import or error) and this change adds an icon to the delete route

    
Changes in this pull request:

-
-
-

@JC5


JC5:
Nice!

PR #9178, opened 2024-08-26
Add  test cases for Api\V1\Controllers\Autocomplete\BillController & BudgetController

tasnim0tantawi:
Add test cases for `Api\V1\Controllers\Autocomplete\BillController.php`

image

---

Add test cases for `Api\V1\Controllers\Autocomplete\BudgetController.php`
image

They are kind of similar, so I felt no need for 2 separate pull requests.

@JC5


JC5:
Nice!

PR #9239, opened 2024-09-14
Fix webhook index page when Firefly is not served at root

jfpedroza:
Changes in this pull request:

When Firefly is not served at the root path, the webhook index page breaks, not listing the webhooks.

image

I added a "." to the URL, like other components do.

Since the solution was simple, I did not open an issue first. 

@JC5


JC5:
Nice, thanks!

PR #9391, opened 2024-10-23
Integer Column

Kaushalbagwe:

Changes in this pull request:

When defining integer columns with ->integer(), it is better to specify the unsigned property for IDs to ensure that they only accept positive values.
Foreign Key Constraints:

You are referencing columns in foreign key constraints that may not exist as intended (especially the transaction_journal_id). Make sure the referenced tables and their columns exist before running the migration.
Decimal Precision:

The decimal precision of balance is set to 32, 12. While this allows for a large number of digits, consider if this is necessary based on the application’s requirements. A more common precision is 10, 2 or 15, 2.
Nullable Foreign Key Columns:

If a foreign key column (like transaction_journal_id) is nullable, make sure the reference table allows for null values in the referenced column.
Unique Constraints:

The unique constraint could potentially allow multiple entries for the same account_id and transaction_currency_id on different dates if they have the same title. If that’s not your intention, then the unique constraint is correctly placed. Otherwise, you may want to reconsider the unique fields.
Nullable Date Column:

You have made the date column nullable, which is fine, but ensure it aligns with the business logic.
 Key Changes Made:
Changed integer to unsignedBigInteger for foreign key columns.
Specified that transaction_journal_id foreign key is nullable.
Adjusted the precision of the balance field to a more common setting of 15, 2.
Ensured the unique constraint is still applied correctly.

@JC5

### DIFF
--- a/app/Enums/AutoBudgetType.php
+++ b/app/Enums/AutoBudgetType.php
@@ -20,16 +20,32 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
-declare(strict_types=1);
+ <?php
 
-namespace FireflyIII\Enums;
-
-/**
- * Enum AutoBudgetType
- */
-enum AutoBudgetType: int
-{
-    case AUTO_BUDGET_RESET    = 1;
-    case AUTO_BUDGET_ROLLOVER = 2;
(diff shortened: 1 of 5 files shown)


JC5:
Hello,

Thanks for your PR. Please let me know what motivated you to submit these changes.

Issue #9458, opened 2024-11-11
The add transaction failed


wqwwffsytxwp:
### Support guidelines

- [x] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)

### I've found a bug and checked that ...

- [x] ... [the documentation](https://docs.firefly-iii.org/) does not mention anything about my problem
- [x] ... there are no open or closed issues that are related to my problem
- [x] ... it's [definitely a Firefly III issue, not me](https://github.com/firefly-iii/firefly-iii/blob/main/.github/its_you_not_me.md)

### Description

Debug information generated at 2024-11-11 20:24:59 for Firefly III version **v6.1.22**.

[...]

PR #9468, opened 2024-11-15
Transaction Model: explicitly cast decimal to string

yparitcher:
Laravel defers to PDO & the underlying database as to what type decimals are cast as. When using sqlite text that match a float will be returned as a float, while mySql always returns a string for a decimal. This leads to crashes with sqlite while trying to import tranansactions. (It does not happen every transaction, only when the balance{before,after} coming from the database is floatable)

`FireflyIII\Support\Models\AccountBalanceCalculator::getLatestBalance(): Return value must be of type string, float returned.`

Fixes: #9458

I figured this out a few days ago, and finally had the time to PR, and saw someone beat me to report the issue. This should be a better & more permanent fix than typecasting the return value, which i only saw you implemented after i opened the pr.

    
Changes in this pull request:

-
-
-

@JC5

### DIFF
--- a/app/Models/Transaction.php
+++ b/app/Models/Transaction.php
@@ -54,6 +54,8 @@ class Transaction extends Model
             'bill_name_encrypted' => 'boolean',
             'reconciled'          => 'boolean',
             'balance_dirty'       => 'boolean',
+            'balance_before'      => 'string'
+            'balance_after'       => 'string'
             'date'                => 'datetime',
         ];
 


JC5:
Nice, that's a better fix indeed.

yparitcher:
On mobile bow but we should probably also revert b0bfb556dbdce70b10e525c2cbbc35a96ade944b

JC5:
Meh, it does little harm.

Issue #6121, opened 2022-06-01
Cannot create trigger rules for field values with multiple spaces


nicklozon:
### Support guidelines

- [X] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)

### I've found a bug and checked that ...

- [X] ... [the documentation](https://docs.firefly-iii.org/) does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem
- [X] ... it's [definitely me, not you](https://github.com/firefly-iii/firefly-iii/blob/main/.github/its_you_not_me.md)

### Description

Rule Triggers do not respect when the field value has multiple spaces. I.e. "Description is.." on value "Meow Mix&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;PAY"

### Expected behaviour

Multiple spaces do not prevent the trigger rule from properly matching on the field.

### Debug information

Debug information generated at 2022-05-31 22:20:23 America/Toronto for Firefly III version **5.7.9**.

| Scope | Version |
| --- | --- |
| Firefly III | 5.7.9 |
| Firefly III API | 1.5.6 |
| Build | 190, 31-05-2022 06:23:12 CEST |
| DB version | 18 (exp. 18) |
| PHP | `8.0.17` |
| Host | `Linux` |

| System info | Value |
| --- | --- |
| System TZ | America/Toronto |
| Browser TZ | America/Toronto |
| App environment | local |
| App debug mode | true |
| App cache driver | file |
| App logging | debug, stack |
| Display errors | Off |
| Error reporting | ALL errors |
| Interface | apache2handler |
| Default language | en_CA |
| Default locale | equal |
| BCscale | 24 |
| DB drivers | mysql, pgsql, sqlite |
| Current driver | mysql |
| Trusted proxies (.env) | `` |

| User info | Value |
| --- | --- |
| Login provider | eloquent  |
| User guard | web  |
| Headers | N/A, N/A  |
| Stateful domains |  |
| Session start | 2022-05-01 00:00:00 |
| Session end | 2022-05-31 23:59:59 |
| Session first | 2022-01-31 00:00:00 |
| User ID | 1 |
| User language | en_US |
| User locale | en_US |
| Attempt at "en_US.utf8" | 'en_US.utf8' |
| Attempt at "en_US.UTF-8" | 'en_US.UTF-8' |
[...]

Issue #9123, opened 2024-08-05
Descriptions with multiple spaces cannot be matched by "description is"


Azelphur:
### Support guidelines

- [X] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)

### I've found a bug and checked that ...

- [X] ... [the documentation](https://docs.firefly-iii.org/) does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem
- [X] ... it's [definitely a Firefly III issue, not me](https://github.com/firefly-iii/firefly-iii/blob/main/.github/its_you_not_me.md)

### Description

It seems like the matching engine has problems with spaces, if you create a transaction with a description that contains multiple spaces, you cannot match it.

### Debug information

Debug information generated at 2024-08-05 19:41:04 for Firefly III version **v6.1.19**.

[...]

Issue #8631, opened 2024-03-05
Rules not triggerd with Tags longer than 36 Characters


sn0west:
### Support guidelines

- [X] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)

### I've found a bug and checked that ...

- [X] ... [the documentation](https://docs.firefly-iii.org/) does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem
- [X] ... it's [definitely a Firefly III issue, not me](https://github.com/firefly-iii/firefly-iii/blob/main/.github/its_you_not_me.md)

### Description

Tags over 36 Character do not trigger rules. 
I've tested it in the demo site as well.

### Debug information

Debug information generated at 2024-03-05 18:30:43 for Firefly III version **v6.1.9**.

[...]

Issue #6170, opened 2022-06-19
Search query with open/close parens does not return known-valid matches


skorvek:
### Support guidelines

- [X] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)

### I've found a bug and checked that ...

- [X] ... [the documentation](https://docs.firefly-iii.org/) does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem
- [X] ... it's [definitely me, not you](https://github.com/firefly-iii/firefly-iii/blob/main/.github/its_you_not_me.md)

### Description

Search in description containing open and close parentheses does not match transactions

### Expected behaviour

Search for description is  or contains or starts with "Requested transfer to SCOTT J KORVEK (MANASQUAN BANK Checking xxxx)" where xxxx is part of my account number returns 0 matches.  Truncating the query to "Requested transfer to SCOTT J KORVEK (MANASQUAN BANK" works  - removing the closing parens is what allows non-zero matches.

### Debug information

Debug information generated at 2022-06-18 21:33:02 America/New_York for Firefly III version **5.7.9**.

| Scope | Version |
| --- | --- |
| Firefly III | 5.7.9 |
| Firefly III API | 1.5.6 |
| Build | 196, 09-06-2022 18:30:26 CEST |
| DB version | 18 (exp. 18) |
| PHP | `8.0.17` |
| Host | `Linux` |

| System info | Value |
| --- | --- |
| System TZ | America/New_York |
| Browser TZ | America/New_York |
| App environment | production |
| App debug mode | true |
| App cache driver | redis |
| App logging | notice, stack |
| Display errors | Off |
| Error reporting | ALL errors |
| Interface | apache2handler |
| Default language | en_US |
| Default locale | equal |
| BCscale | 24 |
| DB drivers | mysql, pgsql, sqlite |
| Current driver | pgsql |
| Trusted proxies (.env) | `**` |

| User info | Value |
| --- | --- |
| Login provider | eloquent  |
| User guard | web  |
| Headers | N/A, N/A  |
| Stateful domains |  |
| Session start | 2022-06-01 00:00:00 |
| Session end | 2022-06-30 23:59:59 |
| Session first | 2001-01-01 00:00:00 |
| User ID | 1 |
| User language | en_US |
| User locale | en_US |
| Attempt at "en_US.utf8" | 'en_US.utf8' |
[...]

PR #9598, opened 2025-01-01
New Query Parser for Search Engine and Rules

Sobuno:
This PR fixes issue #6121, #9123, #8631, #6170, #4938, #1371

Changes in this pull request:
- Add interface `QueryParserInterface` to allow multiple implementations of a QueryParser
- Abstract `Gdbots\QueryParser` behind `GdbotsQueryParser` to conform to the above interface.
- Add unit tests verifying behaviour of QueryParserInterface implementations
- Add additional custom query parser `QueryParser`. As a result, the following changes in Rule matching/Searches compared to the `Gdbots\QueryParser`-based implementation:
  - Multiple sequential spaces are no longer reduced to a single space
  - Search: A minus sign in front of a word now searches for descriptions without the word. Previously, it would search for descriptions including the word thus essentially ignoring the minus sign
  - Search: Subqueries can now contain fields and nested subqueries rather than just words
  - A right parenthesis in the middle of a word no longer results in it being split into two. Example: `a)b` would be parsed as `a b`.
  - Subqueries can now be negated by adding a minus sign in front of the parentheses - This will flip the search for all parameters inside, making subqueries actually useful

## Context
I use the Data Importer to import transactions through GoCardless. A lot of transactions have multiple sequential spaces in their description, meaning the `Create rule based on transaction` link cannot be directly used as evidenced in some of the issues mentioned above. I got curious as to why it was not an easy fix and when I saw the reliance on gdbots/query-parser, I found it a fun challenge for the holidays to write my own parser.

The custom query parser is a single-pass parser meaning each character is only looked at once, resulting in `O(n)` performance in parsing the query.

## Notes on unit tests
The set of unit tests will not pass for the `Gdbots\QueryParser` implementation. This is intentional to show the issues I found while implementing my own parser. The final Pull Request would remove the `Gdbots\QueryParser` reliance entirely.

## Additional notes on whitespace
Displaying the whitespace on the UI is a two-fold issue (and is not fixed by this PR): 
* The CSS `white-space` property should be applied where white-space should be retained
* Twig's [Whitespace Control](https://twig.symfony.com/doc/3.x/templates.html#templates-whitespace-control) shoud be utilized correctly - At the moment, a lot of the indentation that is being done for readability of the Twig templates end up being rendered due to using `{{ something }}` rather than `{{- something -}}`

@JC5


JC5:
Very cool, this looks amazing! Let me know how I can help.

Sobuno:
> Very cool, this looks amazing! Let me know how I can help.

Wow, that was a fast response. I think my main concern is if I've captured all corner-cases as I am rather new to Firefly III. Do you have any advanced search strings we would want to test?

Also, I haven't touched PHP in 10+ years, so I might be a bit rusty - I am having difficulties making the environment-variable based selection for the dependency injection actually work; since the final Pull Request would remove Gdbots entirely, it's not that big of a deal, but it's annoying me.

Sobuno:
I have implemented the negation of subqueries as well now and refactored the tests to make them simpler.

The UI also has gotten a tweak to be able to more clearly see which words are grouped together (very faint grey color around each "word") and to include wording for excluded strings:
image

JC5:
This looks good, very nice! I see a few minor details code-wise but nothing exciting. 

> Do you have any advanced search strings we would want to test?

Not really. There is no support for AND/OR yet because the current search engine just builds massive queries and those are a PITA to combine with AND/OR and sub-queries anyway. A project for the future perhaps. Other than that, if the parser can deal with it, so can the search.

I'll do a final review over the weekend and merge it into develop asap. I'll make sure the copyright notices mention your GitHub handle as well.

Thanks again!

Issue #8948, opened 2024-06-03
To have a rule triggered both on create AND on update


fabianonetto:
### Support guidelines

- [X] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)
- [X] My request is not listed as [a very good idea, but unfortunately...](https://docs.firefly-iii.org/explanation/more-information/what-its-not/)
- [X] I've used [the search](https://github.com/firefly-iii/firefly-iii/issues?q=is%3Aissue) and this has not been requested before.

### Description

- I would like to have the option to create a rule that runs is triggered when a transaction is created AND when it is updated. It only allows to select 1 option as of now

### Solution

To replace the current dropdown with a couple of checkboxes and adjust the rule engine 
![image](https://github.com/firefly-iii/firefly-iii/assets/15900163/1bc9e26a-9da4-44ff-b203-a9ee2d0bbe80)

### What are alternatives?

To create 2 rules with the same conditions and actions with different triggers

JC5:
No, this will stay as it is due to database limitations. Sorry.

PR #9644, opened 2025-01-10
Add ability to set a rule that will trigger both when rules are created or updated

Sobuno:
This PR fixes issue #8948. 

It got annoying to have to decide whether my rules needed to trigger when updating or when creating as I do some sanitizing of data through webhooks - For those cases, I have to wait for the sanitizing update to happen. In other cases, the data is fine up front so no update happens.

Changes in this pull request:

- Add a third option to the Rules dropdown, `store-or-update-journal` (In English named "When a transaction is created or updated")
- Modified all code that looked for `store-journal` or `update-journal` to also look for `store-or-update-journal` when fetching rules
   - To do this, `getRuleGroupsWithRules` now take an optional array rather than an optional string
- Modified all logic that handles creation of rules to also handle this new case.

@JC5

### DIFF
--- a/app/Api/V1/Requests/Models/Rule/StoreRequest.php
+++ b/app/Api/V1/Requests/Models/Rule/StoreRequest.php
@@ -119,7 +119,7 @@ class StoreRequest extends FormRequest
             'description'                => 'min:1|max:32768|nullable',
             'rule_group_id'              => 'belongsToUser:rule_groups|required_without:rule_group_title',
             'rule_group_title'           => 'nullable|min:1|max:255|required_without:rule_group_id|belongsToUser:rule_groups,title',
-            'trigger'                    => 'required|in:store-journal,update-journal,manual-activation',
+            'trigger'                    => 'required|in:store-journal,update-journal,manual-activation,store-or-update-journal',
             'triggers.*.type'            => 'required|in:'.implode(',', $validTriggers),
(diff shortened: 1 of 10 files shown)


JC5:
No, sorry.

The first thing this will do is push the rule engine into an eternal loop. I appreciate the problem, but I'm not interested in a fix for this problem.

Sobuno:
> No, sorry.
> 
> The first thing this will do is push the rule engine into an eternal loop. I appreciate the problem, but I'm not interested in a fix for this problem.

I am not sure I follow - How will this be worse than me creating two separate rules matching on the same content in terms of hitting an infinite loop?

JC5:
No you're right, I don't think it works like that. It used to, though. My bad.

Still, not a fan of the option like it is right now. `store-or-update-journal` doesn't feel right. I'd much rather add multiple triggers to the rule, that feels a lot cleaner than this. But that means many `user-action` searches need improvement.

I'm not going to merge anything new before 6.2 anyway. Do you mind if I take a stab at this later?

Sobuno:
> Still, not a fan of the option like it is right now. `store-or-update-journal` doesn't feel right. I'd much rather add multiple triggers to the rule, that feels a lot cleaner than this.

Agree, this will lead to combinatorial explosion if more event types start appearing - It was just the easy way to go :)

> I'm not going to merge anything new before 6.2 anyway. Do you mind if I take a stab at this later?

Sure, no problem

PR #9743, opened 2025-02-02
Feature nordic currencies

mansehr:
Changes in this pull request:

- Adding SEK
- Adding NOK
- Adding ISK

Most of the nordic countries in Europe have some different flavours of 'Krona'. I only found the Danish krone in the repo so I added the Swedish, Norwegian and Icelandic to be more inclusive. This feels inline with the currency exchange you are currently implementing.

References:
 - https://sv.wikipedia.org/wiki/Svensk_krona
 - https://no.wikipedia.org/wiki/Norsk_krone
 - https://is.wikipedia.org/wiki/%C3%8Dslensk_kr%C3%B3na
@JC5


JC5:
Nice, thanks!

PR #9815, opened 2025-02-11
chore: update translated  with  zh-tw.json only translated 

changchiakai:
This fix is ​​to update the taiwan json translation

tks

@JC5


JC5:
Hey there!

This is very helpful, but the translations are managed externally: https://crowdin.com/project/firefly-iii

Could you submit your changes there? Otherwise, they will be overwritten during the next build.

Thanks in advance for your help.

changchiakai:
OK tks

Issue #9833, opened 2025-02-14
Dokku: vendor/bin/heroku-php-nginx: not found


halilim:
### Support guidelines

- [x] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)

### I've found a bug and checked that ...

- [x] ... [the documentation](https://docs.firefly-iii.org/) does not mention anything about my problem
- [ ] ... there are no open or closed issues that are related to my problem
- [x] ... it's [definitely a Firefly III issue, not me](https://github.com/firefly-iii/firefly-iii/blob/main/.github/its_you_not_me.md)

### Description

> There is no more support in the standard Docker images for linux/arm/v7 and linux/arm/v8 I'm afraid.
> 
> Thank you for pointing me in the right direction! I should read the changelogs, I guess... I might try a non-Dockerfile deployment option. By the way, could you elaborate? I thought the official PHP image still [supports](https://hub.docker.com/_/php#:~:text=Supported%20architectures) it. 

 _Originally posted by @halilim in [#9715](https://github.com/firefly-iii/firefly-iii/issues/9715#issuecomment-2647159317)_

I have created a new AMD VPS from scratch, and it's having the same issue:

```log
$ arch
# => x86_64
$ dokku logs ...
...
[i] Database is already migrated.
...
Thank you for installing Firefly III, v6.2.6!
Firefly III should be ready for use.
...
/usr/local/bin/docker-php-serversideup-entrypoint: 69: exec: vendor/bin/heroku-php-nginx: not found
```

### Debug information

N/A

JC5:
Are you running Firefly III on heroku, by any chance?

halilim:
It was an Oracle compute instance, deployed with `dokku git:from-image app fireflyiii/core:version-6.2.6` ([docs](https://dokku.com/docs/deployment/methods/image/))

JC5:
Well it's not surprising that it's not being found, it's not in the container. But I don't know what `dokku git:from-image app` does and what it does to the image, sorry.

halilim:
I have tried it on a new Hetzner x86_64 VPS and a local Ubuntu VM using Multipass with the same results (all three were Ubuntu 24.04). At this point, the reproduction steps are down to:

```sh
# https://dokku.com/docs/getting-started/installation/#1-install-dokku
wget -NP . https://dokku.com/install/v0.35.15/bootstrap.sh
[...]

PR #9880, opened 2025-02-20
fix: ignore the Procfile when building the docker image

josegonzalez:
This PR fixes issue #9833.

Changes in this pull request:

- This PR removes the Procfile from the built docker image. The Procfile is heroku-specific, causing issues on platforms which respect it when specifying processes to run for docker image or dockerfile based deploys.

@JC5

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+Procfile


JC5:
Nice, merged.

PR #9903, opened 2025-02-24
Add support for PHP function `strpos` in expressions

lompi:
Changes in this pull request:
- Add support for PHP function `strpos` in expressions

@JC5


JC5:
Nice work!

PR #10614, opened 2025-07-18
Reduced minimum password length from 16 to 10 for usability

VijayKumarDVK96:
The minimum required password length has been reduced from 16 to 10 characters.

While longer passwords are more secure, requiring 16 characters proved to be impractical for many users. A 10-character minimum strikes a better balance between security and usability, encouraging users to choose strong but manageable passwords.

This change updates the validation rule accordingly.

    
Changes in this pull request:

-
-
-

### DIFF
--- a/app/Http/Requests/ProfileFormRequest.php
+++ b/app/Http/Requests/ProfileFormRequest.php
@@ -43,7 +43,7 @@ class ProfileFormRequest extends FormRequest
         // fixed
         return [
             'current_password'          => 'required',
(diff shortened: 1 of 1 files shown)


VijayKumarDVK96:
The minimum required password length has been reduced from 16 to 10 characters.

JC5:
Hi,

No. The minimum password length will stay what it is.

https://docs.firefly-iii.org/references/faq/firefly-iii/using/#why-is-the-minimum-password-length-16-characters

PR #10993, opened 2025-10-02
use correct translation key for category report income table

ctrl-f5:
use correct translation key for category report income table. 
The key already exists in the translations, but the incorrect one was used:

```
'spent_average'  => 'Spent (average)',
'income_average' => 'Income (average)',
```


JC5:
Nice find, must be some lazy copy/pasting by yours truly 😉 .

PR #11024, opened 2025-10-07
improved balance range date handling

ctrl-f5:
- new parameter `$inclusive` added to `Steam::accountsBalancesOptimized()`
  - this makes the `$date` filter inclusive/exclusive, optional and default true makes it BC
  - this change negates the need to subtract a second from dates being passed
- new helper function `Steam::accountsBalancesInRange($start, $end)`
  - as there ware a lot of places with 2 seperate calls to `Steam::accountsBalancesOptimized()` to get a start and end balance of a range, this seems like a good one to extract
  - added benefit of having the `$inclusive` parameter handled here, so it can not be forgotten in the future
- removed calls to `Carbon::subSecond()` and `Carbon::endOfDay()` where they were not needed anymore, or moved it to the relevant Request object where the dates are already parsed and validated


JC5:
Nice work, merged!

PR #11039, opened 2025-10-09
proposal for improved request handling

ctrl-f5:
Currently validating, converting, and calculating inputs or related data is a little bit scattered: in the existing request objects, in the base controller, or in the controller actions themselves.
This leads to inconsistent implementations, and bugs concerning the same issue that need to be fixed in different places.

To combat this, I would propose to consolidate all request handling to the Request objects, which seems the most reusable and logical place.

## The basics

Let's take the `PaginationRequest` as an example:

```php
class PaginationRequest extends FormRequest
{
    public function rules(): array
    {
        return [
            'sort'  => ['nullable', new IsValidSortInstruction($this->sortClass)],
            'limit' => 'numeric|min:1|max:131337',
            'page'  => 'numeric|min:1|max:131337',
        ];
    }

    public function withValidator(Validator $validator): void
    {
        $validator->after(
            function (Validator $validator): void {
                $limit = $this->convertInteger('limit');
                if (0 === $limit) {
                    $limit = (int)Preferences::getForUser(auth()->user(), 'listPageSize', 50)->data;
                }

                $page = $this->convertInteger('page');
                $page = min(max(1, $page), 2 ** 16);
                $offset = ($page - 1) * $limit;
                $sort = $this->get('sort');

                $this->attributes->set('limit', $limit);
                $this->attributes->set('sort', $sort);
                $this->attributes->set('page', $page);
                $this->attributes->set('offset', $offset);
            }
        );
    }
}
```

We centralize parsing here:
- leverage the FormRequest to perform the same validation every time
- fetch userDefaults if necessary
- calculate nice to have fields (`$offset` from `$page` and `$limit`, which is now done in a lot of places in controllers)
- All results are set on the `Request::$attributes`

All this logic and mental strain can be removed from the controllers, you know the request will give you valid, usable, data.

In the controller we can leverage deconstructing of `$request->attributes->all()`.

```php
public function index(PaginationRequest $request)
{
    [
        'page' => $page,
        'limit' => $limit,
        'offset' => $offset,
        'sort' => $sort,
    ] = $request->attributes->all();

    $collection  = $this->repository->getAccountsByType($sort);
    $accounts    = $collection->slice($offset, $limit);
}
```

## Composing Requests using `AggregateApiRequest`

Let's take `AccountController::show()` as example, which is the one where we have also implemented the proposel in this PR as a POC:

The request has multiple options: pagination data, a date range, a simple date, and account types.

We have a `ApiRequest` for each of those groups of options. Each of them is usable on their own, but we want to use multiple at the same time. For this we can use `AggregateApiRequest`:

```php
class ShowRequest extends AggregateApiRequest
{
    protected function getRequests(): array
    {
        return [
            [PaginationRequest::class, 'sort_class' => Account::class], // we need to know which model is being sorted
            DateRangeRequest::class,
            DateRequest::class,
            [AccountTypeApiRequest::class, 'required'], // we could make these fields required
        ];
    }
}
```
This will combine all the rules and validators of all given Request classes and execute them accordingly.

Additionaly, if you pass an array instead of a request class, those will be passed into the Requests so they can adapt their behavior accordingly, or simple because they require additional information to function correctly.

### next steps

Again, this is just a proposal to improve things, I'm happy to adapt, extend, or simplify whatever you deem necessary. I'm pretty sure you'll have to live longer with this than me, as I'm just a contributor.

The current implementation for those 1 or 2 actions is as good as mergable imho. And if it gets green lit, I'm happy to spend some time converting existing controllers.

### DIFF
--- a/app/Http/Controllers/Rule/SelectController.php
+++ b/app/Http/Controllers/Rule/SelectController.php
@@ -24,6 +24,7 @@ declare(strict_types=1);
 
 namespace FireflyIII\Http\Controllers\Rule;
 
+use Carbon\Carbon;
 use FireflyIII\Exceptions\FireflyException;
 use FireflyIII\Http\Controllers\Controller;
 use FireflyIII\Http\Requests\SelectTransactionsRequest;
@@ -56,7 +57,7 @@ class SelectController extends Controller
 
         $this->middleware(
             static function ($request, $next) {
-                app('view')->share('title', (string) trans('firefly.rules'));
+                app('view')->share('title', (string)trans('firefly.rules'));
                 app('view')->share('mainTitleIcon', 'fa-random');
 
                 return $next($request);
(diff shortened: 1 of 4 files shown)


JC5:
Hey, thanks for this, I really appreciate it. The code looks good, I will go over it during the weekend. I like the idea and the architecture, I would have never thought of this myself.

Issue #9447, opened 2024-11-08
Transaction doesn't show up when attaching HTML file


Marc928132:
### Support guidelines

- [x] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)
- [x] My request is not listed as [a very good idea, but unfortunately...](https://docs.firefly-iii.org/explanation/more-information/what-its-not/)
- [x] I've used [the search](https://github.com/firefly-iii/firefly-iii/issues?q=is%3Aissue) and this has not been requested before.

### Description

When attaching a HTML file to a withdrawal, said withdrawal doesn't show up in the "All expenses between..." list under Accounting>Transaction>Expenses.
I does show up at (all?) other places like the dashboard or the charts under Accounting>Transaction>Expenses.
When opening the transaction from one of these other places like the dashboard the attached HTML file is not there.

Firefly III version 6.1.21 via Docker on Unraid accessed via Firefox 132.0.1

### Solution

-

JC5:
Hi, thanks for opening an issue.

Right now, `text/html` files are not allowed for uploading. I'll pick this up for the next release, and make sure you get a nice error if you upload anything forbidden. 

Also, the issue of not showing the transactions will be fixed.

Thanks for reporting!

Issue #6332, opened 2022-08-08
Attachment upload fails (unallowed MIME type application/octet-stream)


Kaemmelot:
### Support guidelines

- [X] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)

### I've found a bug and checked that ...

- [X] ... [the documentation](https://docs.firefly-iii.org/) does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem
- [X] ... it's [definitely me, not you](https://github.com/firefly-iii/firefly-iii/blob/main/.github/its_you_not_me.md)

### Description

When uploading a certain PDF file to an existing transaction (the file extension and name don't matter) I get a success message, but the attachment is not visible in the transaction itself. When accessing `/attachments` the file is shown as 0 KB and not present. I found a hidden error on the `/debug` page but I don't know where this error originates from.

### Expected behaviour

I would at least expect a visible message about this error. At best the file upload should just work.

### Debug information

Debug information generated at 2022-08-08 19:29:56 Europe/Berlin for Firefly III version **5.7.10**.

| Scope | Version |
| --- | --- |
| Firefly III | 5.7.10 |
| Firefly III API | 1.5.6 |
| Build | (unknown), (unknown) |
| DB version | 18 (exp. 18) |
| PHP | `8.1.4` |
| Host | `Linux` |

| System info | Value |
| --- | --- |
| System TZ | Europe/Berlin |
| Browser TZ | Europe/Berlin |
| App environment | local |
| App debug mode | false |
| App cache driver | file |
| App logging | notice, stack |
| Display errors | Off |
| Error reporting | ALL errors |
| Interface | fpm-fcgi |
| Default language | de_DE |
| Default locale | equal |
| BCscale | 24 |
| DB drivers | mysql |
| Current driver | mysql |
| Trusted proxies (.env) | `` |

| User info | Value |
| --- | --- |
| Login provider | eloquent  |
| User guard | web  |
| Headers | N/A, N/A  |
| Stateful domains |  |
| Session start | 2022-08-01 00:00:00 |
| Session end | 2022-08-31 23:59:59 |
| Session first | 2022-02-01 00:00:00 |
| User ID | 1 |
| User language | de_DE |
| User locale | de_DE |
| Attempt at "de_DE.utf8" | 'de_DE.utf8' |
| Attempt at "de_DE.UTF-8" | 'de_DE.UTF-8' |
[...]

PR #11041, opened 2025-10-09
Add XML mimetypes to the allowedMimes list

jreyesr:
Hello, and thanks for all the work on Firefly! 
    
This PR adds support for XML files as attachments, which currently fails with a 422 error because those files aren't in the list of allowed filetypes. XML appears to be a somewhat common choice for electronic invoices, which if [these](https://cloud.esker.com/fm/einvoicing-compliance-map/) [lists](https://www.billtrust.com/products/automated-invoicing-software/einvoicing-compliance-map) are to be trusted are already used in the EU and a good portion of Latin America. Thus, I'd expect adding support for XML invoices to be useful to people in multiple countries, for storing the invoice of each transaction

Changes in this pull request:

- Add `text/xml` and `application/xml` as allowed MIME types for attachments. The XMLs that I tested with are detected as `text/xml` so that appears to be what PHP's automatic file type detection prefers, but apparently [there's an esoteric difference](https://www.grauw.nl/blog/entry/489/) regarding implicit encodings, so I added both just in case

Alternatively, perhaps it's better to allow people to add additional MIME types via an environment var? In the past three years there have been four additions of MIME types that I could find: #10846, https://github.com/firefly-iii/firefly-iii/commit/ca1b048758a92a6aba2289146ce30e42b0bd278a, #9447 and #6332. If you'd think an envvar is a better choice than adding Yet Another MIME Type, I can do that too

### DIFF
--- a/app/Factory/PiggyBankEventFactory.php
+++ b/app/Factory/PiggyBankEventFactory.php
@@ -26,6 +26,7 @@ namespace FireflyIII\Factory;
 use FireflyIII\Models\PiggyBank;
 use FireflyIII\Models\TransactionJournal;
 use FireflyIII\Repositories\PiggyBank\PiggyBankRepositoryInterface;
+use Illuminate\Support\Facades\Log;
 
 /**
  * Create piggy bank events.


JC5:
Nice, merged!

PR #11052, opened 2025-10-11
correct validator function to check for errors + `Account\ShowControllerTest`

ctrl-f5:
Significant bug in how I check for validation errors :)

That was also the reason the `DateRequest` required an idditional setter before validator was checked.

FIgured out my local test setup and added a Controller test to ensure it works now.


JC5:
Nice work, merged!

PR #11414, opened 2025-12-28
Add Uruguayan Peso to currency seeder

oboxodo:
Adds Uruguayan Peso so it doesn't need to be manually added on new installs and (hopefully) also starts getting regular automated exchange rates.


JC5:
Nice, merged. Exhange rates should be coming up soon.

Issue #11560, opened 2026-01-19
Transaction list overflows parent container when description is long


gian21391:
### Support guidelines

- [x] I've read the  [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)

### I've found a bug and checked that ...

- [x] ... [the documentation](https://docs.firefly-iii.org/) does not mention anything about my problem
- [x] ... there are no open or closed issues that are related to my problem
- [x] ... it's [definitely a Firefly III issue, not me](https://github.com/firefly-iii/firefly-iii/blob/main/.github/its_you_not_me.md)

### Description

When a transaction has a very long description, the table containing the transaction list expands beyond the width of its parent container (.box-body). This causes the layout to break, pushing other columns (like Amount, Date, Actions) off-screen.

### Expected Behavior:

The transaction list table should remain within the bounds of the layout. The description column should wrap text or truncate with an ellipsis, or the container should provide a scrollbar to handle the overflow gracefully without breaking the page layout.

### Actual Behavior:

<img width="1461" height="713" alt="Image" src="https://github.com/user-attachments/assets/03dc4247-87a2-408b-b233-bdbea2ff091d" />

### Debug information

Debug information generated at 2026-01-20 00:47:05 for Firefly III version **v6.4.16**.

[...]

PR #11569, opened 2026-01-21
Fix layout overflow issues with long content in v1 and v2 layouts

gian21391:
This PR fixes issue #11560.
This ensures that containers scroll horizontally instead of breaking the layout when displaying very long transaction descriptions.

Changes in this pull request:

- Added overflow-x: auto to .box-body, .box, and .content-header classes in the v1 CSS (firefly.css).
- Added overflow-x: auto to the same classes in the v2 SCSS (app.scss).

### DIFF
--- a/public/v1/css/firefly.css
+++ b/public/v1/css/firefly.css
@@ -263,6 +263,10 @@ span.twitter-typeahead {
     background-color: #f5f5f5;
 }
 
+.box-body, .box, .content-header {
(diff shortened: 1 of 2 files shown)


JC5:
Nice, merged!

JC5:
I'm going to reverse this for the time being. It also pushes menu's under the layout, for example when viewing a transaction.

Issue #7620, opened 2023-06-09
Issues with light mode


rchl:
### Support guidelines

- [X] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)

### I've found a bug and checked that ...

- [X] ... [the documentation](https://docs.firefly-iii.org/) does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem
- [X] ... it's [definitely a Firefly III issue, not me](https://github.com/firefly-iii/firefly-iii/blob/main/.github/its_you_not_me.md)

### Description

There are issues with colors of some form elements when light mode is selected in firefly preferences but dark mode is preferred system-wide.

### Debug information

Debug information generated at 2023-06-09 22:42:24 Europe/Oslo for Firefly III version **6.0.11**.

| Scope | Version |
| --- | --- |
| Firefly III | 6.0.11 |
| Firefly III API | 2.0.3 |
| Build | 490, 27-05-2023 07:21:11 CEST |
| Base Build | 3, 29-04-2023 14:45:54 CEST |
| DB version | 19 (exp. 19) |
| PHP | `8.2.5` |
| Host | `Linux` |

| System info | Value |
| --- | --- |
| System TZ | Europe/Oslo |
| Browser TZ | Europe/Oslo |
| App environment | local |
| App debug mode | false |
| App cache driver | file |
| App logging | notice, stack |
| Display errors | Off |
| Error reporting | ALL errors |
| Max upload | 67108864 (64 MB) |
| Interface | apache2handler |
| Default language | en_US |
| Default locale | nb_NO |
| BCscale | 12 |
| DB drivers | mysql, pgsql, sqlite |
| Current driver | mysql |
| Trusted proxies (.env) | `` |

| User info | Value |
| --- | --- |
| Login provider | eloquent  |
| User guard | web  |
| Headers | N/A, N/A  |
| Stateful domains |  |
| Session start | 2023-06-01 00:00:00 |
| Session end | 2023-06-30 23:59:59 |
| Session first | 2022-03-01 00:00:00 |
| User ID | 1 |
| User language | en_GB |
| User locale | en_GB |
| Attempt at "en_GB.utf8" | 'en_GB.utf8' |
| Attempt at "en_GB.UTF-8" | 'en_GB.UTF-8' |
[...]

Issue #8613, opened 2024-03-03
Some minor color issues


rumpff:
### Support guidelines

- [X] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)

### I've found a bug and checked that ...

- [X] ... [the documentation](https://docs.firefly-iii.org/) does not mention anything about my problem
- [X] ... there are no open or closed issues that are related to my problem
- [X] ... it's [definitely a Firefly III issue, not me](https://github.com/firefly-iii/firefly-iii/blob/main/.github/its_you_not_me.md)

### Description

I have found a few theme specific color issues, for each of them the system theme was set to dark

### Debug information

Debug information generated at 2024-03-03 05:57:41 for Firefly III version **v6.1.9**.

[...]

PR #11589, opened 2026-01-25
apply user-selected light/dark mode to form elements (checkboxes, date picker) #8613 #7620

mateuszkulapl:
This PR fixes issue
#7620 - [unresolved part of this - checkboxes](https://github.com/firefly-iii/firefly-iii/issues/7620#issuecomment-1585620341)
also related to #8613 (point 7)

Changes in this pull request:
- Set meta tag color-scheme based on user dark mode preference

Previously, if the `<meta name="color-scheme">` was set to `light dark`, browsers would render form elements (e.g., checkboxes, date pickers) according to the system's color scheme. This caused inconsistencies when the user selected "Always light" or "Always dark" in Firefly III but the system theme differed. With this update, form elements now respect the user's selected theme regardless of system preferences.

Examples:
Left: Firefly III preference light, system dark
Right: Firefly III preference dark, system light

Before:
before-dark-checkboxes
After:
before-dark-checkboxes-after


JC5:
Ohh very nice, thank you!

PR #11632, opened 2026-02-01
fix v2 layout dashboard transactions load

mateuszkulapl:
Changes in this pull request:
- update transactions function to accept multiple query params. Previously params were incorrectly nested under 'page' param like
```
page[page]  => 1
page[start] => ...
page[end]  => ...
 ```
which caused error and broke account list and transaction display on the dashboard


JC5:
Ah nice, thanks. The v2 layout needs love again :)

PR #11713, opened 2026-02-14
Add "Download selected" CSV export option to transaction bulk actions

barreeeiroo:
Include a "Download selected" button in the transactions list, to allow downloading the selected transactions into a CSV file:

image

Sample CSV: [2026_02_14_selected_transactions.csv](https://github.com/user-attachments/files/25316779/2026_02_14_selected_transactions.csv)

---

**Disclaimer**: Code modifications in this PR were assisted by Claude Opus 4.6, using Claude Code. I did review and tested all the code manually.


barreeeiroo:
I was not sure about whether to place the button above or below the edit ones... My preference would be above the Edit ones, but wanted to hear opinions before making it the "first" one in there...

JC5:
Hey,

I'm sorry, I can't merge this. I would have appreciated a heads-up as not to disappoint you.

1. There is already a CSV generator for transactions, so now there is double code to deal with. Plus, they both have different keys and values. 
2. I don't want to add new features to the current layout that are not originating, or also available over the API. Because now I need to add another API endpoint for the same feature, resulting in double code.
3. I have no requests for this feature and I would rather not add this without a clear ask and goal.

As a whole, I like the idea of being able to export  specific of transactions using the API, but nobody's asking me to build it so it's really not on the top of my list.

barreeeiroo:
Hey @JC5, no worries!

1. Yeah, I know there's an "Export" functionality, but that was too broad and I couldn't really say "I want to download the ones for X account". That being said, I'd be fine to wire up this code with that CSV generator.
2. Is the V2 layout functional enough as of now? I'm happy to take care of adding support for this there as well, in case that's the problem. I understand it would increase maintainability issues, but I'd say a feature to export/download data with more fine-grained filters would be something useful in general.
3. Can I open an issue as a feature request? I actually faced a need to download the transactions to verify some sums, and typing the values manually was too tedious. I was mainly looking at the Search results page, which gave me exactly the values, but I needed to operate on them so downloading to CSV/Excel was way more convenient.

JC5:
Hey,

Glad to hear it. The current exporter is just a "mass" exporter right now, and may need some love as well at some point. Depending on your needs, you could also hack something together using the API.

Feel free to open an issue regardless. The v2 layout is not on par yet, but I'm currently cleaning up middleware and authentication routes so there's more stability to fix that up again. 

Cheers,

James

PR #12038, opened 2026-03-26
Fix: ensure today's transactions are included in dashboard balance

giuscia:
When viewing the dashboard for an interval ending in the future (like 'This month'), the balance filter limits the query up to 'now()'. This strictly excludes any newly created transactions for 'today' if they are saved with an end-of-day time (or affected by timezone offsets).

This fix forces the limit strictly to the end of the day, ensuring everything from today is always included properly in the dynamic calculation.

Assisted-by: Gemini via Antigravity

I'm ten years burnin' down the road
Nowhere to run, ain't got nowhere to go

### DIFF
--- a/app/Support/Twig/General.php
+++ b/app/Support/Twig/General.php
@@ -152,8 +152,8 @@ class General extends AbstractExtension
             $session          = clone session('end', today(config('app.timezone'))->endOfMonth());
             if ($session->lt($date)) {
                 $date = $session->copy();
-                $date->endOfDay();
             }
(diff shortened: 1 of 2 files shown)


JC5:
Hi,

This is intentional and will not be changed, thanks for your PR.

Cheers,  
James

giuscia:
"Thanks for the feedback, James! My bad—it was a total oversight on my part, I had forgotten to set the correct TZ. Cheers!"

PR #12072, opened 2026-04-01
Format transaction date using moment to adjust timezone

### LINKED ISSUE #12062: Date is incorrect when editing a transaction
### Support guidelines

- [x] I'm smart and I read the  [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)

### I've found a bug and checked that ...

- [x] ... [the documentation](https://docs.firefly-iii.org/) does not mention anything about my problem
- [x] ... there are no open or closed issues that are related to my problem
- [x] ... it's [definitely a Firefly III issue, not me](https://github.com/firefly-iii/firefly-iii/blob/main/.github/its_you_not_me.md)

### Description

The time in the date field is different when editing.

View page:
Image

Edit page:
Image

This happened in both my self-hosted setup and the demo site (https://demo.firefly-iii.org/transactions/show/433).

### Debug information

N/A

### Steps to reproduce

1. Create a transaction.
2. View the transaction and check the time.
3. Edit the transaction and see that the time is different.

### Additional info

_No response_

zyrouge:
#### Reference issues and PRs

Fixes #12062

Instead of using the UTC timestamp, the date-time is adjusted to local timezone and then set.

#### AI usage disclosure

I used AI assistance for:
- [ ] Code generation (e.g., when writing an implementation or fixing a bug)

#### Any other comments?

This was not testing by running locally. I debugged it in the browser. Let me know if I should make any changes.

image

@JC5

### DIFF
--- a/resources/assets/v1/src/components/transactions/EditTransaction.vue
+++ b/resources/assets/v1/src/components/transactions/EditTransaction.vue
@@ -469,10 +469,12 @@ export default {
             if (typeof window.expectedSourceTypes === 'undefined') {
                 console.error('window.expectedSourceTypes is unexpectedly empty.')
             }
+            let parsedDate = moment.parseZone(transaction.date);
             let result = {
                 transaction_journal_id: transaction.transaction_journal_id,
                 description: transaction.description,
-                date: moment(transaction.date).format().substring(0, 16),
+                dateOffsetTarget: parsedDate.utcOffset(),
+                date: parsedDate.local().format().substring(0, 16),
                 reconciled: transaction.reconciled,
                 amount: this.roundNumber(this.positiveAmount(transaction.amount), transaction.currency_decimal_places),
                 category: transaction.category_name,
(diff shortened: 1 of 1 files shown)


JC5:
It look like that if you do this, every time you edit the transaction, even without changing the date, the time and date of the transaction will jump X hours. Where X is the difference between the system time zone and the browser time zone.

So three edits to fix the description, if your browser = Europe/London and your system = Europe/Amsterdam, the transaction will have magically jumped 3 hours.

zyrouge:
That fix seemed to work on the demo instance when editing and saving. But, in demo instance, the date seems to be saved incorrectly even when creating a transaction. I have not idea how date is parsed in the backend, but even when sending the date-time along timezone does seem not work. I think handling this could fix it.
If this does not involve major changes, I could look into this if you give me pointers on where the data is handled in the backend.

JC5:
Sorry but no, nobody is playing around with that right now.

zyrouge:
Well, that's a bummer. I think this should be looked into since this seems to occur during transaction creation. In my instance, the date-time is only wrong when editing. In the demo, the date-time is wrong when creating and editing. As this patch fixes the issue in my instance where my system and the server is in the same timezone, can we go ahead and merge it?

JC5:
No. Feel free to patch your own instance though.

zyrouge:
@JC5 Can you check with these changes? Solution looks ugly, but this should handle the time offset.

zyrouge:
Bumping this.

JC5:
Sorry, completely fell of my plate. I'll check it out!

JC5:
I'm going to leave this as it is right now. Sorry about that.

Issue #12176, opened 2026-04-23
[Feature] Password security check on client side for better UX


tasnim0tantawi:
### Support guidelines

- [x] I've read the [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)
- [x] My request is not listed as [a very good idea, but unfortunately...](https://docs.firefly-iii.org/explanation/more-information/what-its-not/)
- [x] I've used [the search](https://github.com/firefly-iii/firefly-iii/issues?q=is%3Aissue) and this has not been requested before.

### Description

<img width="593" height="690" alt="Image" src="https://github.com/user-attachments/assets/f435486d-3628-4760-b521-0870fc4b0d8e" />

We need to submit the register request in order to get the password security violations. Would be nicer if we have it at client side so input does not get erased.

### Solution

Client side validation in the frontend. @JC5 what do you think? Will work on it if it's okay!

JC5:
Hey, thanks for opening an issue.

Of course, should be easy enough. `minlength=16` should do it!


Cheers,
James

PR #12179, opened 2026-04-23
implement password validation JS script

tasnim0tantawi:
Fixes: https://github.com/firefly-iii/firefly-iii/issues/12176

Now, we can see client side errors when registering with password that fails to meet requirements:
image

If having one issue:
image

Please note I have used AI to help me implement this feature, made sure that it follows same code style as login blade and have reviewed it myself.
 Did the https://api.pwnedpasswords.com/range/ api request on client side.
#### AI usage disclosure

I used AI assistance for:
- [x] Code generation (e.g., when writing an implementation or fixing a bug)

@JC5

### DIFF
--- a/app/Http/Middleware/SecureHeaders.php
+++ b/app/Http/Middleware/SecureHeaders.php
@@ -63,7 +63,7 @@ class SecureHeaders
             "base-uri 'self'",
             // "form-action 'self'", // safe
             "font-src 'self' data:",
-            sprintf("connect-src 'self' %s", $trackingScriptSrc),
+            sprintf("connect-src 'self' https://api.pwnedpasswords.com %s", $trackingScriptSrc),
             sprintf("img-src 'self' data: 'nonce-%1s' ", $nonce),
             "manifest-src 'self'",
         ];
--- a/resources/views/auth/register.blade.php
+++ b/resources/views/auth/register.blade.php
@@ -21,6 +21,10 @@
         </div>
     @endif
 
+    <div id="client-errors" class="alert alert-danger" role="alert" style="display:none;">
+        <ul id="client-errors-list"></ul>
+    </div>
+
     <div class="card">
         <div class="card-body register-card-body">
             <p class="login-box-msg">{{ trans('firefly.register_new_account') }}</p>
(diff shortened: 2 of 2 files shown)


JC5:
Hey, thanks for your PR.

Nice work!

vmario89:
Hi. I understand the feature, but what is https://pwnedpasswords.com? https://api.pwnedpasswords.com ist not made for humans, as it is the API endpoint. There is no service /website available to see some vendor/imprint which offers an info page. I dislike external services which are not trushtworty on a glance.

I would appreciate to switch over to something official like https://haveibeenpwned.com/api/v3

niklas2810:
> Hi. I understand the feature, but what is https://pwnedpasswords.com? https://api.pwnedpasswords.com ist not made for humans, as it is the API endpoint. There is no service /website available to see some vendor/imprint which offers an info page. I dislike external services which are not trushtworty on a glance.
> 
> I would appreciate to switch over to something official like https://haveibeenpwned.com/api/v3

Okay, so I wasn't the only one being freaked out for a second there 😅 A more detailed communication/documentation (in the PR description) would have been great, especially since this is one of the few cases where Firefly III "talks to the outside world".

Turns out everything is fine here though IMO. The actual password checks for haveibeenpwned run on a different domain, this is even mentioned in the official docs for this endpoint: https://haveibeenpwned.com/api/v3#SearchingPwnedPasswordsByRange 

Also, I've found this blog article. Seems like Cloudflare at least supports them to provide the API, maybe that's why it's running under a different domain: https://blog.cloudflare.com/validating-leaked-passwords-with-k-anonymity/

JC5:
Fair enough, I should have clarified that. I will do so in the info box.

JC5:
Just to be clear:

- The JS and the internal password security (the client side first, then the server side) check both use the same anonymous API (https://haveibeenpwned.com/API/v3#PwnedPasswords). 
- Both checks are opt-in as of Firefly III v6.6.2 and later. For earlier versions, this is op-out. The registration form uses a checkbox. If you do not check it, no data will be sent.
- The check is done using a clever hash of your password, that does not allow your password to be restored or deduced from it.

PR #12182, opened 2026-04-24
fix shrinked sidebar expanding when navigating by clicking on icons

tasnim0tantawi:

Saving collapsed sidebar state in local storage and restoring it when a new page we navigate to loads.

Apr-24-2026 15-52-15

#### AI usage disclosure

I used AI assistance for:
- [ ] Code generation (e.g., when writing an implementation or fixing a bug)

#### Any other comments?

@JC5

### DIFF
--- a/public/v1/js/ff/firefly.js
+++ b/public/v1/js/ff/firefly.js
@@ -70,13 +70,6 @@ $(function () {
         }, 0);
     });
 
-    (function () {
-        try {
-            if (localStorage.getItem('ff3_sidebar_collapsed') === '1') {
-                document.body.classList.add('sidebar-collapse');
-            }
-        } catch (_) {}
-    })();
 
 
     // on submit of form, disable any button in form:
--- a/public/v1/js/ff/firefly.js
+++ b/public/v1/js/ff/firefly.js
@@ -56,6 +56,20 @@ $(function () {
         return false;
     });
 
+    // save sidebar collapsed state when page loads.
+    $(document).on('click', '[data-toggle="push-menu"]', function () {
+        setTimeout(function () {
+            try {
+                if (window.localStorage) {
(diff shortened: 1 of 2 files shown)


JC5:
You could implement this in `/v1/js/ff/firefly.js` and use the jquery code as well, perhaps?

tasnim0tantawi:
> You could implement this in `/v1/js/ff/firefly.js` and use the jquery code as well, perhaps?

Done! Please check
Apr-24-2026 21-20-07

JC5:
That would also imply all of the code fits in firefly.js. You've told me you're using AI tools to get this done, I would like to ask you to see what you can do without them.

tasnim0tantawi:
> That would also imply all of the code fits in firefly.js. You've told me you're using AI tools to get this done, I would like to ask you to see what you can do without them.

Did not get you :( I'm using inline AI (autocomplete) so I went to the file and wrote there. Do you mean you want me to write it all in firefly.js? @JC5  Will remove the code from default.twig

tasnim0tantawi:
Okay no AI :( 

When moving the code to firefly.js or any external script the sidebar has a small delay until the collapse class applies, GIF cannot capture it :( @JC5 When researching, it's because inline script is executed before page rendering in default twig. Lmk what do u think. I can try use CSS to fix it

tasnim0tantawi:
image
Having this in default twig has no delay and class gets applied right away. @JC5 Should I keep it separated or all in default.twig?

JC5:
That's not the reason it loads right away. When you put it in `firefly.js` it becomes part of `$(function () {` which waits for the document to load.

JC5:
No you're right, inline is the fastest. Using `(function () {` is faster, but inline is the fastest.

PR #12519, opened 2026-07-22
fix(firefly-iii): correct budget currency calculations 🍌🍌🍌

petrov-vladyslav:
- Match expenses by primary or foreign currency and convert all relevant transactions to the primary currency

- Correct budget limit API and front-page chart aggregation

- Add an environment switch that restores legacy behavior

- Add integration coverage and ignore local IDE and PHPUnit artifacts

Assisted-by: GPT-5 via Codex

#### AI usage disclosure

I used AI assistance for:
- [x] Code generation (e.g., when writing an implementation or fixing a bug)
- [x] Test/benchmark generation
- [x] Documentation (including examples)
- [x] Research and understanding

#### Any other comments?

@JC5

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -293,6 +293,11 @@ DEMO_PASSWORD=
 #
 FIREFLY_III_LAYOUT=v1
 
+# Set this to true to restore the legacy budget calculation behavior. Legacy
+# calculations only match a transaction's primary currency to the budget
+# currency and ignore foreign currency amounts.
+USE_LEGACY_BUDGET_CURRENCY_BEHAVIOR=false
+
 #
 # Please make sure this URL matches the external URL of your Firefly III installation.
 # It is used to validate specific requests and to generate URLs in emails.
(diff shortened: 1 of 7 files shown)


JC5:
Hi,

Thanks for your effort, but I will not be able to merge this.

1. I do not want "legacy" behavior. The system works one way. The discussion on how to deal with this behavior affects more places than what you've touched on here.
2. https://docs.firefly-iii.org/explanation/contributing/#no-ai-contributions

petrov-vladyslav:
Hi @JC5,

I understand. I’m still not sure why this behavior couldn’t be made optional, but that’s fine — I’ll roll back to the previous version.

JC5:
Optional also means I have to keep up with both versions, so double the maintenance.

petrov-vladyslav:
That’s up to you, of course. However, this new behavior creates a real problem: it becomes impossible to properly track a budget when using multiple currencies. Setting the transaction’s foreign currency to the budget’s native currency does not solve the problem.

Issue #11531, opened 2026-01-16
Performance: updateRunningBalance executes even when use_running_balance is disabled, causing timeouts on Mass Edits


maxime-killinger:
### Support guidelines

- [x] I've read the  [support guidelines](https://github.com/firefly-iii/firefly-iii/blob/main/.github/support.md)

### I've found a bug and checked that ...

- [x] ... [the documentation](https://docs.firefly-iii.org/) does not mention anything about my problem
- [x] ... there are no open or closed issues that are related to my problem
- [x] ... it's [definitely a Firefly III issue, not me](https://github.com/firefly-iii/firefly-iii/blob/main/.github/its_you_not_me.md)

### Description

I am experiencing severe performance issues (504 Gateway Timeouts) when performing Mass Edits on transactions. My database contains significant history (~23k transactions).
I have explicitly disabled the "Running Balance" feature to improve performance, but profiling reveals that `UpdatedGroupEventHandler::updateRunningBalance` is still being executed for every modified transaction, taking ~25 seconds per transaction to complete.

It appears the event handler code does not check the `use_running_balance` configuration before launching the heavy recalculation process.

### Debug information

Debug information generated at 2026-01-16 07:15:52 for Firefly III version **v6.4.15**.

[...]

PR #12624, opened 2026-08-17
Default USE_RUNNING_BALANCE to false (disabled), matching the GUI text

### LINKED ISSUE #12623: USE_RUNNING_BALANCE defaults to true, contradicting the GUI text ("disabled by default") and a "not used" comment
### Description

Three things contradict each other about the `USE_RUNNING_BALANCE` env var:

1. `config/firefly.php` (line 78):

```php
'running_balance_column' => (bool)env_default_when_empty(env('USE_RUNNING_BALANCE'), true), // this is only the default value, is not used.
```

The fallback value is `true`.

2. The GUI settings page says (lang key `setting_use_running_balance_explain`, en_US):

> "It is still slightly experimental, so disabled by default."

3. The inline comment claims the default value "is not used" — but it **is** used as the fallback in `AppConfiguration::get('use_running_balance', config('firefly.feature_flags.running_balance_column'))`.

### Consequence

Because neither `.env.example` nor the official docker-compose template contains a `USE_RUNNING_BALANCE` line, **every fresh deployment that follows the docs gets the running-balance feature enabled by default**. Creating a transaction then triggers a full-database running-balance recalculation.

On my deployment (Firefly III v6.6.6, Docker, ~3,300 transactions being imported with the Data Importer) each transaction took ~8 seconds, i.e. several hours for the whole import, with heavy database load. After disabling the "running balance" feature in the GUI, the remaining import ran at under a second per transaction.

### Suggested fix

Change the fallback to `false` so the code matches the GUI text ("experimental, disabled by default") — or, if `true` is intended, update the GUI text and mention the env var in `.env.example`. Either way, the "is not used" comment is misleading and should be fixed.

### Related

- #11531 ("Performance: recalculation is executed even after disabling") was fixed recently and shows this feature has known performance issues.

jazzzzmaybe:
## What

`USE_RUNNING_BALANCE` currently defaults to **true**, which contradicts the GUI settings text that says the running-balance feature is "slightly experimental, so **disabled by default**". Because neither `.env.example` nor the docker-compose template contains a `USE_RUNNING_BALANCE` line, every fresh deployment that follows the docs silently enables the feature — and creating a transaction then triggers a full-database running-balance recalculation (measured ~8s per transaction during a large import; under a second after disabling it).

This PR:

1. `config/firefly.php` — changes the fallback from `true` to `false`, matching the GUI text, and fixes the misleading `// this is only the default value, is not used.` comment: the value **is** used as the fallback by `AppConfiguration::get()` at several call sites.
2. `app/Api/V1/Controllers/System/ConfigurationController.php` — uses the same config fallback as the other six call sites instead of a hardcoded `true` (otherwise the API would still report `use_running_balance=true` on fresh installs).
3. `.env.example` — documents `USE_RUNNING_BALANCE` so the toggle is discoverable.

## Why

- Fixes #12623
- Related: #11531 (running-balance recalculation was still executed after disabling — fixed recently), which shows the feature has known performance issues and should stay opt-in.

## Verification

- No tests reference `running_balance`, so no test changes are required.
- Behavior: on a fresh install (no `USE_RUNNING_BALANCE` set, no stored configuration) `config('firefly.feature_flags.running_balance_column')` now returns `false`; the GUI checkbox, the API response and the actual recalculation behavior now agree. Existing installations with the value stored in the database are unaffected.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -299,3 +299,9 @@ FIREFLY_III_LAYOUT=v1
 # It is used to validate specific requests and to generate URLs in emails.
 #
 APP_URL=http://localhost
+
+#
+# The running balance feature is experimental and disabled by default.
+# Set to true to enable it. You may need to restart your container afterwards.
+#
+USE_RUNNING_BALANCE=false
--- a/app/Api/V1/Controllers/System/ConfigurationController.php
+++ b/app/Api/V1/Controllers/System/ConfigurationController.php
@@ -153,7 +153,7 @@ final class ConfigurationController extends Controller
         $singleUser            = AppConfiguration::get('single_user_mode', true);
         $lastCheck             = AppConfiguration::get('last_update_check', 1);
         $enableExchangeRates   = AppConfiguration::get('enable_exchange_rates', config('cer.enabled'));
-        $useRunningBalance     = AppConfiguration::get('use_running_balance', true);
+        $useRunningBalance     = AppConfiguration::get('use_running_balance', config('firefly.feature_flags.running_balance_column'));
         $enableExternalMap     = AppConfiguration::get('enable_external_map', false);
         $enableExternalRates   = AppConfiguration::get('enable_external_rates', false);
         $allowWebhooks         = AppConfiguration::get('allow_webhooks', false);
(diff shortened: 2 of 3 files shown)


jazzzzmaybe:
Superseded by #12625, which targets the `develop` branch as required. Apologies for the noise.

PR #12696, opened 2026-09-04
Restyle Settings > Users, date-range picker, and add Light/Dark/System theme switcher

SudarshanTAPMI:
## Summary
- **Settings > Users**: colored icon badges on each section header (Invite/All users/All invited users), status-dot pills for the admin/2FA/blocked columns, and row actions moved to a labeled right-hand column with separated icon buttons — matching the card/table language established elsewhere in the app.
- **Date-range picker**: retheme the header's date-range control (and the Reports page's own picker) onto the app's existing design tokens instead of its vendored default/dark/light CSS — mono uppercase headers, accent-tint range highlighting, solid accent circles for selected dates, and an outline ring (not a fill) for "today".
- **Theme switcher**: replace the header's single click-to-flip light/dark icon with a 3-way Light/Dark/System segmented control, reusing the same three choices and translation strings already offered on the Preferences page. `PreferencesController::setDarkMode()` was previously hard-restricted to `light`/`dark` only ("binary by design") — relaxed to accept all three modes via `config('firefly.available_dark_modes')`.
- Fixed the new theme switcher rendering flush against the header's top edge instead of vertically centered with its sibling nav items.
- Added an `--ip-accent-secondary` token pair (light + dark) for the purple icon badge, following the existing `--ip-info`/`--ip-positive` pattern.

## Test plan
- [x] Settings > Users verified live in both light and dark mode — icon badges, status dots, and right-aligned row actions all render correctly
- [x] Date-range picker verified live in both light and dark mode on the Budgets page header (ranges list, calendar grid, in-range/selected/today states, Apply/Cancel buttons)
- [x] Theme switcher: all three segments (Light/Dark/System) tested live — each correctly persists via `setDarkMode()` and reloads into the right theme; active segment highlight moves correctly
- [x] Theme switcher vertical alignment fixed and reverified against sibling header items (search box, date range, "Hey Jude" menu)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


JC5:
Hi there!

This is an automated reply. `Share and enjoy`

You triggered an automated reply, because it seems you removed or changed the AI assistance disclosure from the PR template. Without a valid disclosure, your PR cannot be processed.

Even if you did not use AI, this disclosure must be present. Please reply to your PR and explain your use of AI in any or all of the following areas:

1. Code generation (e.g., when writing an implementation or fixing a bug).
2. Test/benchmark generation
3. Documentation (including examples)
4. Research and understanding

There cannot be interaction with your PR without this disclosure.

If the disclosure is present but the bot did not pick up on it, please accept my apologies for the intrusion. Contrary to other bots, this one is just a simple `bash` script and it may be wrong.

SudarshanTAPMI:
Apologies — this was opened in error by an automated tool (a `gh pr create` call defaulted to this upstream repo instead of the intended fork). No action needed here; it's already closed. Sorry for the noise.

